PR #437, opened 2014-07-30
Do not use Transformer for searchable text indexer

puittenbroek:
Forcing the output to text/plain can lead to following Exception:
No transform path found from 'text/x-html-safe' to 'text/plain'. 

Leading to nothing being indexed for the specific tile that causes the error.

The catalog index takes care of the output we provide. Cleaning it up beforehand is not needed.
Ancient Plone (Products.Archetypes.BaseObject -> SearchAbleText) works similar.


hvelarde:
any news on this?

puittenbroek:
Haven't had time to look into this again, refreshing my local build now. Hope to spend some time on this tomorrow. The problem is that the transformer we use now does not always work. But if you provide html, the catalog will transform it for us. So your test failing makes sense, since I would expect html output :)

puittenbroek:
@hvelarde Hi Hector, I added a new commit with two extra assertions to clarify the problem I'm trying to resolve. The RichTextValue you created didn't reflect how we create them in richtext tiles, I updated this.

I'm doing a manual transform on the two richtextvalue's to show the problem that's currently present.
We output 'text/x-html-safe' and try to transform that into 'text/plain' for the index, which fails, resulting in an empty output.

hvelarde:
ok, great!

puittenbroek:
pypi failed on one of the checks it seems

src/collective/cover/content.py
```diff
@@ -149,12 +149,21 @@ def assign_id_for_tiles(cover, event):
 def searchableText(obj):
     """Return searchable text to be used as indexer. Includes id, title,
     description and text from Rich Text tiles."""
-    transformer = ITransformer(obj)
     tiles_text = ''
+    transforms = getToolByName(obj, 'portal_transforms', None)
```
hvelarde: use plone.api here: ``api.portal.get_tool('portal_transforms')``

src/collective/cover/content.py
```diff
@@ -149,12 +149,21 @@ def assign_id_for_tiles(cover, event):
...
     for t in obj.list_tiles('collective.cover.richtext'):
         tile = obj.restrictedTraverse(
             '@@collective.cover.richtext/{0}'.format(str(t)))
-        tiles_text += transformer(tile.data['text'], 'text/plain')
+        value = tile.data['text']
+        data = transforms.convertTo('text/plain',
```
hvelarde: follow conventions on [breaking lines](http://docs.plone.org/external/plone.api/docs/contribute/conventions.html#breaking-lines)

src/collective/cover/tests/test_cover.py
```diff
@@ -94,6 +94,48 @@ def test_layoutmanager_settings(self):
...
+                              "type": "tile"}],
+                "data": {"column-size": 8, "layout-type": "column"},
+                "id": "group1",
+                "roles": ["Manager"],
+                "type": "group"}],
+  "class": "row",
+  "type": "row"}]
+"""
+        tile = self.cover.restrictedTraverse('collective.cover.richtext/test1')
+        value1 = RichTextValue(raw=u'<p>01234</p>',
```
hvelarde: follow conventions on [breaking lines](http://docs.plone.org/external/plone.api/docs/contribute/conventions.html#breaking-lines)

src/collective/cover/tests/test_cover.py
```diff
@@ -94,6 +94,48 @@ def test_layoutmanager_settings(self):
...
+                               mimeType='text/x-html-safe',
+                               outputMimeType='text/x-html-safe')
+        data_mgr = ITileDataManager(tile)
+        data_mgr.set({'text': value1})
+        tile = self.cover.restrictedTraverse('collective.cover.richtext/test2')
+        data_mgr = ITileDataManager(tile)
+        value2 = RichTextValue(raw=u'<p>56789</p>',
```
hvelarde: follow conventions on [breaking lines](http://docs.plone.org/external/plone.api/docs/contribute/conventions.html#breaking-lines)

puittenbroek:
@hvelarde I'll update tomorrow morning, it's on my work laptop :)

hvelarde:
thank you! I don't have a strong opinion on the transform; for me, the tests are passing and seem to be ok; the only weird think is the that the text has 2 spaces between the rich text stuff but I think that doesn't make any difference.

if you want, have someone else to take a look on it; I'll merge it as soon as you finish and mr. Travis gives me green light.

BTW, don forget to add a line on top of the changelog describing the fix.

puittenbroek:
I'm not sure what causes the failed test..?

puittenbroek:
@hvelarde All seems to be working, I have no-one available for second opinion. Except for the unwanted extra spaces, I think that works as desired. From my point of view at least :)

PR #455, opened 2014-10-03
Make package compatible with jQuery 1.9

### LINKED ISSUE #413: Dragging in list tile only works once.
Initially reported via e-mail by @hvelarde to me and @fredvd .  Now confirmed.

Pinging @naro because this is broken since his commit 268b394c45af098a715c389cc4740410c810c2b8. That commit is for compatibility with jQuery 1.9+, but it has negative effects for jQuery 1.7 that Plone 4.3 uses.
- Create a cover with a list tile containing two (or more) items.
- Go to the compose tab.
- Drag one of the items to a different spot in the list tile. This works, but now you cannot drag items anymore. You need to load the Compose tab again to get this back.

@naro, can you look at this?

### LINKED ISSUE #498: Incompatible with jQuery 1.9+ (pin to 1.8.3?)
As collective.cover uses the ``$.live`` method (maybe other things too), it's incompatible with jQuery 1.9+.  plone.app.jquery should be pinned in this setup.py to be prevent install with an incompatible version.

Somewhat related to #492 regarding pinning.

hvelarde:
fixes #413 
closes #498

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.0a12 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
+- Make package compatible with jQuery 1.9.
+  [naro]
+
 - Fix issue where page composition layer had side-effects outside cover pages by upgrading plone.app.blocks to 2.0 (fixes `#331`_);
   drop support for Plone 4.2.
   [datakurre]
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   matrix:
     - PLONE_VERSION=4.3
     - PLONE_VERSION=4.3 JQUERY=1.8.3
+    - PLONE_VERSION=4.3 JQUERY=1.9.1.3
     - PLONE_VERSION=5.0
 matrix:
   allow_failures:
--- a/src/collective/cover/static/compose.js
+++ b/src/collective/cover/static/compose.js
@@ -1,7 +1,7 @@
 
 (function ($) {
     $.fn.liveSortable = function (opts) {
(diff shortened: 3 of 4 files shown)


hvelarde:
so, are we talking about Plone 4.3.5 then?

hvelarde:
@rodfersou review this and let me know if we can merge it; tests are passing.

FYI @cewing @cguardia @alecpm anything else?

CHANGES.rst
```diff
@@ -228,6 +228,10 @@ There's a frood who really knows where his towel is.
 - Fix for generic title for list tiles introduced in 1.0a7 (fixes `#393`_).
   [fredvd]
 
+- Use jQuery's .on instead of deprecated .live method. This allows
```
hvelarde: this should be moved to the latest release and rewritten to something like "Make package compatible with jQuery/1.9."

rodfersou:
@hvelarde this is the right way to structure live selectors now, we bind the selector to document (almost root of our html), this way even if the element didn't exist it will be associated with jquery event after creation.

davilima6:
@rodfersou For performance reasons it's best to bind to nearest non-mutable ancestor. See: 

* http://stackoverflow.com/questions/16383718/are-there-any-drawbacks-to-listen-events-on-document-level-in-jquery
* http://stackoverflow.com/questions/12824549/should-all-jquery-events-be-bound-to-document

rodfersou:
@davilima6 thank you for your commend, I didn't know about this practice, and it makes all sense for cases where optimization is needed.  I'm going to make some tests here.

hvelarde:
@davilima6 would you mind to open another issue to check that? this is only to gain compatibility with jQuery 1.9 and is not related with performance issues in any way.

PR #456, opened 2014-10-03
Carousel tile with cycle2

### LINKED ISSUE #442: Replace Galleria with Cycle2 on carousel tile?
after months of using it, we have arrive to the conclusion that [Galleria](http://galleria.io/) is huge and difficult to theme.

some weeks ago we discovered [Cycle2](http://jquery.malsup.com/cycle2/) and we are very pleased with what we saw; we even created a new [Photo Gallery](https://github.com/simplesconsultoria/sc.photogallery) add-on to test it and found it full-featured, really versatile and easy to theme.

Should we replace **Galleria** with **Cycle2**?

djowett:
Hi, I'd like some good thoughts on this, though I should first mention outstanding items (which should be implemented next week) are:
   * Swipe gesture - (Cycle2 needs "jQuery Mobile" to get this working - is this included in the dependencies somewhere?)
   * Correct template to use all configured fields (as in list.pt)

That aside, I'd particularly like thoughts on:
* whether switching between galleria & cycle2 themes could be simpler
* whether doc amendments are in the right place & say too little or too much
* if test fixture set up in testing.py is a good way to do this
* should Cycle2 view look (as currently) fairly close to galleria view, or be distinct

Or anything else...

closes #442


hvelarde:
closing this as we have a new package providing this feature: https://github.com/collective/covertile.cycle2

PR #470, opened 2014-12-18
display a collection in a Bootstrap carousel

### LINKED ISSUE #338: Transform collections and list tiles in Slideshow
We should create a easy way to transform an individual collection tile or list tile into a slideshow. This could replace the complicated Caroussel tile in a more elegant way.

That can be easily implemented using one of this simple and small slideshow scripts:
1. http://unslider.com/
2. http://responsiveslides.com/

The main ideia is:
1. Go to layout tab;
2. Click in collection tile configuration link;
3. Choose the CSS class "Slideshow";
4. Save;
5. Go to view tab and your collection tile is presented as a slideshow.

Its connected to #225 because we need to transform collection tile and list tiles in real HTML list.

Some CSS styles shoul be needed to enhance the presentation.

ebrehault:
This PR allows to display collection items in a Bootstrap carousel.
Unlike Galleria (used in the existing Carousel tile) , the Bootstrap carousel is not restricted to images only, it can show any HTML content as slides.
That's the only reason why I made this change.
Nevertheless it might be quite confusing for users if 2 different tiles offer a carousel feature in 2 diferent ways...
So I am not sure it must be merged, but it is a way to launch a discussion about it.

closes #338


hvelarde:
meanwhile, could you please update your pull request? code analysis is failing

marcosamm:
I needed to create a tile to make the display of a collection as carousel on the homepage of the company. The collection sorts and filters the news and the tile displays the photo and title of each news. I called him Collection Carousel.
The basic idea of my solution was to create a class with the implementation based on CollectionTile (inheritance) and visualization based on carousel.pt.
It is important to have a component that displays from a collection, displaying the standard for object image and allowing change the setting to display HTML content as slide.

ebrehault:
@hvelarde Hi Hector, I have merged the last changes from master.

hvelarde:
for me this makes a lot of sense: we can remove completely the carousel tile from the core, diminishing the complexity of the package and the number of dependencies, and adding the possibility of having a simple carousel for lists and collections OOTB.

on the other side we can let open the possibility of adding a carousel tile from outside, with as many options as needed, by using whatever library you may want.

the only risk I can foresee here is that we must provide a clear path for migration:

* by providing a carousel tile based on Galleria that you can just install and make all carousel tiles still usable
* or, just by migrating any instance of a carousel tile to a list with the new options set up.

and this could lead to broken layouts.

@runyaga @frapell @fredvd @mauritsvanrees @polyester @rodfersou comments?

polyester:
I would be all in favour of having the List and Collection tiles be 'carouselable', and would not worry *too* much about migration. In all honesty, I've had all carousels break on various upgrades of collective.cover already, one more time is not going to matter that much. They were usually quick to fix manually. As long as the change is documented, I'm all for it.

My preferred option would therefore be:

- migrate them to a list with the new options set up
- present the site admin with the list of URL's of all carousel tiles found, so she or he can make manual amends. 

The biggest problem was always finding out where on the site those pesky editors had added yet another carousel tile and then forgotten about them, so if this second step is possible in some log, that will mitigate all upgrade worries for me.

PR #478, opened 2015-02-04
enable compatibility with jQuery's "noConflict" mode in contentchooser.js

FelixSchwarz:
The "coveractions" object avoids the issue by using only "jQuery" (instead of "$") and I was not sure if the symbol is intentionally global so I didn't include it in the new anonymous function.

### DIFF
--- a/src/collective/cover/static/contentchooser.js
+++ b/src/collective/cover/static/contentchooser.js
@@ -1,3 +1,4 @@
+(function ($) {
 var ajaxSearchRequest = [];
 function contentSearchFilter(url) {
   var queryVal = $("#contentchooser-content-search-input").val();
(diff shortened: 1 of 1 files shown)


FelixSchwarz:
just for the record: I don't think that my code actually broke the build – instead the Travis build failed first after commit 2f8b3fa9e134705da7133d479ff1a0059bf38aba (https://travis-ci.org/collective/collective.cover/builds/45150093)

hvelarde:
the issue with the tests is quite weird; that commit has nothing to do with the error and there were no changes on versions of Selenium or Firefox; I'm trying to fix them now

hvelarde:
tests are fixed now; please update you pull request

Issue #475, opened 2015-01-28
Removing items from existing carousel tiles does not work


fredvd:
In a site upgraded to the latest c.cover master version we can no longer remove carousel items from the carousel compose screen.  Clicking the (x) marker on an item works, but saving the form blinks the following message with "wrong contained type" for a second and then closes without removing the item. The added items are Link CT objects.  
 
![wrong_contained_type](https://cloud.githubusercontent.com/assets/394784/5939383/81f2eefc-a703-11e4-95c6-50ac0cee64c4.gif)

Removing the carousel tile as a whole and re-adding the same Link CT objects solves the problem, then I can remove the same items.

With the updates to the list'y and carousel tiles in the recent months and the double migration, does anyone have an idea what's going on here? @frapell @hvelarde @mauritsvanrees ? 

I've added pdb to the UUIDSFieldDataConverter(BaseDataConverter) class in carousel.py to verify what's going in and out to the form and the data looks sane to mee, all dict()'s. The uuids storage on an existing and migrated carousel tile and the newly created carousel tile also look the same, dicts with the same structure afaics. "Wrong contained type" seems to come from z3c.form (it's also presented as a validation error) when I search. And unfortunately there's no exception/backtrace to look into.

My workaround is to rebuild the carousel tiles in the content db, but when upgrading c.cover to the next released version other users might be affected by this as well.

fredvd:
After more debugging and testing, my workaround to remove the carousel tile and repopulate does not work, the problem pops up on new carousel tiles as well.

The problem seems to be that under some conditions after modifying values in the compose edit for a carousel tile, the browser sends back a json string which is unicode in the request and cannot be parsed by the server. Note the `{u"{u` at the beginning of the value catched in the UUIDSFieldDataConverter ToFieldValue method in carousel.py : 

[...]

Issue #476, opened 2015-02-03
carousel tile stores portal_url in all dropped carousel items when you modify one of the items in the carousel dialog


fredvd:
When you add items to a carousel bry dropping them on the carousel the links and al data that the carousel shows is pulled from the referenced/dropped object by storing the objects UUID

As soon as you go into the edit dialog of the carousel you will see title, description, and url of each item to modify/override. However: as soon as you do this for any single item, all 'references' in the carousel are cut and the data is copied to the carousel: including the full path to the item including the website address. 

This breaks setups where a different domain/backend is used for editing (cms.website.com) or when you set up a new site on a staging domain and afterwards switch the site to another domain to go 'live'.

frapell:
@fredvd I have just tested this and cannot reproduce. This is the internal structure just after dropping 2 objects to a carousel tile using latest master (6692b522ba97292d09c0d308917084cd9b94bb94)

```
{'autoplay': True,
 'count': None,
 'css_class': None,
 'date': None,
 'description': None,
 'image': None,
 'more_link': None,
 'more_link_text': None,
 'tile_title': None,
 'title': None,
 'uuids': {'6df3ab2a7f484cc499fa60cc8bdf50c7': {u'order': u'0'},
           'eaa9dd40b86443f1ae98b8bd8b4b437a': {u'order': u'1'}}}
```

And this is right after manually editing title, description and URL for them
```
{'autoplay': True,
 'count': None,
 'css_class': None,
 'date': None,
 'description': None,
 'image': None,
 'more_link': None,
 'more_link_text': None,
 'tile_title': None,
 'title': None,
 'uuids': {u'6df3ab2a7f484cc499fa60cc8bdf50c7': {u'custom_description': u'This is a description',
                                                 u'custom_title': u'Custom title 1',
                                                 u'custom_url': u'http://www.google.com/custom_1',
                                                 u'order': u'0'},
           u'eaa9dd40b86443f1ae98b8bd8b4b437a': {u'custom_description': u'This is another description',
                                                 u'custom_title': u'Custom title 2',
                                                 u'custom_url': u'http://www.google.com/custom_1',
                                                 u'order': u'1'}}}
```

fredvd:
@frapell And what is the internal structure if you drop 2-3 objects onto an empty carousel, you edit only the title of the first object and then save?

frapell:
[...]

PR #479, opened 2015-02-10
Issue 476 - Fix to check if custom attributes were changed from default value

rodfersou:


rodfersou:
@hvelarde the tests are passing locally

hvelarde:
@fredvd could you please check if this fix your problem?

CHANGES.rst
```diff
@@ -25,6 +25,9 @@ There's a frood who really knows where his towel is.
 - Allow to choose a custom title and description on items in a carousel tile (closes `#459`_).
   [rodfersou]
 
+- Fix to check if custom attributes were changed from default value (closes `#476`_).
```
hvelarde: changelog entries on top of the old ones, please

hvelarde:
I know, but they must pass also remotely; what's the purpose of CI if not ;-)

fredvd:
@hvelarde @rodfersou  Coincedentally this  pull request seems to fix issue #475 . I cannot reproduce 
that one. 

Concerning #476 for which this pull request was made: it fixes it partially, as long as there is plain ASCII in all the fields. Custom fields are still overwritten on the first save when the text contains unicode, so there is I guess something missing to handle unicode on the attribute change check. Here's an example uuids dict where I changed an O into an ö to the tekst on the Link Item that I dropped on the carousel:  Notice the 'xf6' in the unicode string. 


{'count': None, 'description': None, 'title': None, 'css_class': None, 'image': None, 'more_link_text': None, 'more_link': None, 'uuids': {u'a51e533497e149938b93984763a57f61': {u'order': u'0'}, u'bc2b473a5c85406d8ab10b40d5723bf4': {u'custom_description': u'\xf6penbare besturen banen al jarenlang de weg naar pesticidenvrij beheer. Vanaf januari gaan we een stap verder en moeten ook scholen, ziekenhuizen, kinderdagverblijven. pesticidenvrij beheerd worden.', u'order': u'1'}}, 'date': None, 'autoplay': True, 'tile_title': u'Custom Title'}

frapell:
I haven't read the code, but for reading the title of the pull, I wonder... does this handle a case where the user customizes values for added elements to the carousel, and then decides to "go back"? 
In my code, I assumed if field was blank, then I should use the object's data, else, if had any value, use that... Is a case like this considered in this code?

rodfersou:
@frapell I know that your code assumed that the field was blank.. but this is not the default behaviour of other tiles, so I think this is the right way to go.

hvelarde:
@rodfersou please fix the tests so we can merge this

hvelarde:
@fredvd can you please confirm that the issue is fixed?

PR #490, opened 2015-03-16
May have fixed randomly failing Locked Cover test

djowett:
This was causing the following error when running locally, which seems to be related to a [robotframework bug regarding use of imported variables in variable table](https://code.google.com/p/robotframework/issues/detail?id=561). 

    13/17 (76.5%)[ ERROR ] Error in file '.../collective.cover/src/collective/cover/tests/test_locked_cover.robot' in table 'Variables': Setting variable '${ALT_PLONE_URL}' failed: Non-existing variable '${ZOPE_PORT}'.

Fixed for me by moving definitions. So maybe Travis & my set up are both on older versions of robotframework?  My versions are:

    robotframework = 2.7.6
    robotframework-selenium2library = 1.1.0

### DIFF
--- a/src/collective/cover/tests/cover.robot
+++ b/src/collective/cover/tests/cover.robot
@@ -8,6 +8,8 @@ Variables  plone/app/testing/interfaces.py
 ${PORT} =  55001
 ${ZOPE_URL} =  http://localhost:${PORT}
 ${PLONE_URL} =  ${ZOPE_URL}/plone
+${ALT_ZOPE_HOST} =  127.0.0.1
+${ALT_PLONE_URL} =  http://${ALT_ZOPE_HOST}:${PORT}/plone
 ${BROWSER} =  Firefox
 
 ${title_selector} =  input#form-widgets-IBasic-title
--- a/src/collective/cover/tests/test_locked_cover.robot
+++ b/src/collective/cover/tests/test_locked_cover.robot
@@ -9,8 +9,6 @@ Suite Teardown  Close all browsers
 
 *** Variables ***
 
-${ALT_ZOPE_HOST}  127.0.0.1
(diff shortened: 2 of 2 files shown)


hvelarde:
I think the test is failing at the very beginning, when the user tries to log in; that's why I asked this: http://stackoverflow.com/q/27430323/644075

see the RF log here: http://collective-cover.s3-website-sa-east-1.amazonaws.com/artifacts/1752/1752.2/robot_log.html#s1-s16

djowett:
Ok, one more try (not really what I should be spending time on).... I noticed that the only use of "Log in as site owner" (or "Log in" in general is in this test, so maybe there's some issue with that.
I added the Click Link to add a wait after log in in case it's a timing issue.

Just to reiterate - this test is working on my dev setup first with these versions:

    selenium = 2.45.0
    robotframework-selenium2library = 1.5.0
    robotframework = 2.7.6
    plone.app.robotframework = 0.9.8
    robotsuite = 1.2.1

And also working with upgrading these to match versions.cfg from plone.app.robotframework 
    plone.app.robotframework = 0.9.8
    robotframework = 2.8.4
    robotframework-selenium2library = 1.5.0
    robotframework-selenium2screenshots = 0.5.0
    robotsuite = 1.6.1
    selenium = 2.45.0

djowett:
@hvelarde ... hmmm I don't know if that worked or not... (what's the diff between "pr" & "push" in this context?)

I should mention though that I noticed that **both browsers** in my local test were running on localhost:55001.  It seems that one of these two lines (around line 43) redirects the second browser to localhost.

    Enable Autologin as  Site Administrator
    Goto Homepage

It's probably the later (see: http://docs.plone.org/external/plone.app.robotframework/docs/source/plone-keywords/index.html#misc)

hvelarde:
that's exactly the problem; I have no time neither to deal with this but we can left this branch open to try to fix it later.

hvelarde:
test is no longer failing; I'm going to close this.

PR #491, opened 2015-03-17
Set classes for rows and columns

### LINKED ISSUE #284: classes for rows and maybe columns
thinking of fancy one pagers ...
it would be nice to allow rows and maybe columns to have custom attributes that are displayed as class on the element, so someone could style them not only as the ":nth-child(N)" element of a page but style them as "headerteaser", "moodblock", "contentfooter" and so on ...

djowett:
This is done by setting data['view-classes'] in the layout in registry.xml, e.g.:

    <element key="Layout A">[{"type": "row", "class": "row", "data": {"view-classes": "carousel"}, "children": ....

obviously it would be nice to enhance this to allow users to edit the class through the Layout view, that's one for later though. :-)

closes #284


rodfersou:
@djowett I undestand that with current CSS selectors existent it is not possible to style a row looking at it's content (maybe if nothing change in CSS4 spec it will be possible with subject selector http://www.w3.org/TR/selectors4/#subject , but now it is not possible).

So I think this is a nice idea, but I miss an option to change the class when I drop a new row at layout edit..

Correct me if I'm wrong, but the way it is now we will be able to change the class just when create a new default layout, if we remove the row and add another in the place, we will lose the class configuration.

Is it right?

@hvelarde @agnogueira what do you think about this row configuration option to add a class? maybe it is a nice idea to make it for columns too..

rodfersou:
I studied more about CSS4 selectors, and in the editor draft specification it will change the way to do this selection http://dev.w3.org/csswg/selectors-4/#has-pseudo (not sure if it will change again, but one think for sure is that this kind of selector will be present)

djowett:
@rodfersou 

> Correct me if I'm wrong, but the way it is now we will be able to change the class just when create > a new default layout, if we remove the row and add another in the place, we will lose the class configuration.

> Is it right?

Yes, that's correct... at the moment, it should work for me as I'm being quite strict with my client about not changing the layout, and since it isn't breaking it for anyone else a merge at this point would be nice :-)
but as I said it could also be enhanced as you suggest - I can have a brief look, but not a top priority for me.

djowett:
@rodfersou  
Re your points on CSS, this should obviously help theming, but my thoughts were that it would help Diazo too. However, (correct me if I'M wrong) but I believe Diazo is not full CSS 3, so it prob won't get CSS 4 anytime soon

rodfersou:
@djowett you are right.. css4 is far from here.. but about Diazo, you can use xpath for that right now, take a look in ancestor axis http://www.w3schools.com/xpath/xpath_axes.asp

rodfersou:
@djowett maybe you can change the row class just with Diazo and xpath

djowett:
@rodfersou 
No the point is, I need/prefer to *know* which row is which, without guessing as follows:

            <replace
                css:theme-children="#demographics_row"
                css:content="#content .cc-row:nth-child(2)"
                />
I might be able to "make do", but things like just telling which layout is in use (when I have a multi-layout system) are quite tricky at the moment...

djowett:
JS is not my normal game, but could [this jsfiddle](http://jsfiddle.net/gM3ps/43/) be a step in the right direction?

@rodfersou @agnogueira @hvelarde

rodfersou:
@djowett yes, the basic idea would be something like this JS (just need to ensure this data will be saved in the row info).
But that's why I think @hvelarde and @agnogueira should give us their opinion thinking about user experience and layout (maybe an icon and overlay like column size is today should be a better solution)

djowett:
Agreed, I need the UX input.

djowett:
@agnogueira would you mind adding your thoughts on this? I think this would really help my Diazo / CSS situation...

djowett:
@hvelarde @agnogueira I think this is done.  I'd appreciate your feedback

src/collective/cover/layout.py
```diff
@@ -233,6 +242,112 @@ def update(self):
                 i += 1
 
 
+# Ok we exclude a little too much here. We exclude "-89", which is invalid, but
+# we also exclude "-abc" which IS a valid class, but it's also a silly CSS
+# class, so why bother to support it!?
+VALID_CSS_RE = re.compile('[a-zA-Z_][0-9a-zA-Z_-]*')
```
rodfersou: Is it for CSS structure or for class names?

src/collective/cover/layout.py
```diff
@@ -233,6 +242,112 @@ def update(self):
...
+
+def cssConstraint(value):
```
rodfersou: Follow PEP8 in the name of the function https://www.python.org/dev/peps/pep-0008/#function-names

src/collective/cover/layout.py
```diff
@@ -233,6 +242,112 @@ def update(self):
...
+    grok.context(ICover)
+    grok.name('configure-row')
+    grok.require('cmf.ModifyPortalContent')
+
+    schema = IRowConfigureForm
+    ignoreContext = True
+    name = 'configure_row'
+
+
+    def updateWidgets(self):
```
rodfersou: Follow PEP8 for the name of methods https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

src/collective/cover/layout.py
```diff
@@ -233,6 +242,112 @@ def update(self):
...
+                existing_row_data = layout[rownum-1].get('data')
+                if existing_row_data:
+                    self.widgets["row_classes"].value = existing_row_data['view-classes']
+            except IndexError, ie:
+                # Bad rownum parameter, so redirect
+                self.request.response.redirect(self.context.absolute_url())
+
+
+    @button.buttonAndHandler(u"Cancel")
+    def handleCancel(self, action):
```
rodfersou: Follow PEP8 for the name of methods https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

src/collective/cover/layout.py
```diff
@@ -233,6 +242,112 @@ def update(self):
...
+        """User cancelled. Redirect back to the layout view.
+        """
+        layoutView = self.context.absolute_url() + "/@@layoutedit"
+        self.request.response.redirect(layoutView)
+
+
+    @button.buttonAndHandler(u'Save')
+    def handleApply(self, action):
```
rodfersou: Follow PEP8 for the name of methods https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

src/collective/cover/layout.py
rodfersou: maybe VALID_CSS_CLASS_RE is a better name

src/collective/cover/layout.py
djowett: Thanks Rodgrigo for the review.

I agree with everything - but I think here you are refering to 
"updateWidgets", am I correct?
If so I can't change as I'm overriding the superclass defined in z3c.form

Cheers,
Daniel


On 14/04/15 18:01, Rodrigo Ferreira de Souza wrote:
>
> In src/collective/cover/layout.py 
> <https://github.com/collective/collective.cover/pull/491#discussion_r28346739>:
>
> > +
> > +
> > +class RowConfigureForm(form.SchemaForm):
> > +    """ Configure settings for Cover Row
> > +
> > +    """
> > +    grok.context(ICover)
> > +    grok.name('configure-row')
> > +    grok.require('cmf.ModifyPortalContent')
> > +
> > +    schema = IRowConfigureForm
> > +    ignoreContext = True
> > +    name = 'configure_row'
> > +
> > +
> > +    def updateWidgets(self):
>
> Follow PEP8 for the name of methods 
> https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/collective/collective.cover/pull/491/files#r28346739>.
>

-- 
Daniel Jowett

*Jowett Enterprises Ltd*
Tel: 0333 3 553 773
Mobile: 07870 667 126
daniel@jowettenterprises.com
www.jowettenterprises.com

rodfersou: oh got it!

hvelarde:
@djowett we will work on this for now on to include also the settings for columns; thanks for your help!

djowett:
@hvelarde great! Good to know it gets a thorough review. I think maybe the prepoverlay could be done with something quicker than a "noform reload".
  I'm on other stuff for a few days now anyway.

rodfersou:
we are using a different approach, trying to keep thing simple, so this issue will be closes with this pull request: https://github.com/collective/collective.cover/pull/513

PR #499, opened 2015-04-02
Restrict collective.js.jqueryui >=1.9

davidjb:
collective.cover requires at least the curCSS function from 1.9+, it breaks with jqueryui 1.8.x.


hvelarde:
Plone is not compatible with jQuery 1.9 neither as you can see in #413 and #455.

Please, use the [issue_420 branch](https://github.com/collective/collective.cover/tree/issue_420) to work on that issue.

PR #501, opened 2015-04-10
Issue 493 - Performance issue when searching for recent items

rodfersou:
* [x] we should wait for the user to stop typing before making a search (check Plone's [livesearch script](https://github.com/plone/Products.CMFPlone/blob/4.3.x/Products/CMFPlone/skins/plone_ecmascript/livesearch.js))
* [x] search should be done only on the title index
* [x] an spinner should be shown to let the user know there is a search in progress
* [x] we should display the number of search results
* [x] we should return up to 20 elements instead of just 10; batch loading should be by 20 also
* [x] we should show the item description in the title attribute
* [x] we should use the ``grab`` cursor when selecting an item and ``grabbing`` when dragging it

Plus: added Plone default navigation (shows just when there are more than 20 items - Plone default behaviour)

### DIFF
--- a/src/collective/cover/browser/contentchooser.py
+++ b/src/collective/cover/browser/contentchooser.py
@@ -71,7 +71,7 @@ class ContentSearch(grok.View):
     def update(self):
         self.query = self.request.get('q', None)
         self.tab = self.request.get('tab', None)
-        b_size = int(self.request.get('b_size', 10))
+        b_size = int(self.request.get('b_size', 20))
         page = int(self.request.get('page', 0))
         strategy = SitemapNavtreeStrategy(self.context)
 
@@ -99,7 +99,7 @@ class ContentSearch(grok.View):
         catalog_query['portal_type'] = searchable_types
 
         if query:
-            catalog_query = {'SearchableText': u'{0}*'.format(safe_unicode(query))}
+            catalog_query = {'Title': u'{0}*'.format(safe_unicode(query))}
 
         # XXX: not implemented, this is needed?
         # if uids:
--- a/src/collective/cover/browser/contentchooser.py
+++ b/src/collective/cover/browser/contentchooser.py
@@ -100,17 +100,9 @@ class ContentSearch(grok.View):
         # temporary we'll only list published elements
         catalog_query = {'sort_on': 'effective', 'sort_order': 'descending'}
         catalog_query['portal_type'] = searchable_types
-
-        query = safe_unicode(query)
-        results = catalog(**catalog_query)
-
-        # let's filter using python
         if query:
-            results = [
-                b for b in results
-                if (query.lower() in b['Title'].lower() or
-                    query.lower() in b['id'].lower())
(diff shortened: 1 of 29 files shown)


src/collective/cover/static/contentchooser.js
```diff
@@ -1,18 +1,40 @@
 (function ($) {
     var ajaxSearchRequest = [];
-    function contentSearchFilter(url) {
+    var timeoutIDs = [];
+    function contentSearchFilter(b_start) {
```
rodfersou: sorry @hvelarde I missed your pré-review of this code, and don't remember what should I change here..

src/collective/cover/browser/contentchooser.py
```diff
@@ -98,16 +101,17 @@ def search(self, query=None, page=0, b_size=10, uids=None):
...
         results = catalog(**catalog_query)
-        results = Batch(results, size=b_size, start=(page * b_size), orphan=0)
 
+        # let's filter using python
+        if query:
+            results = [
+                b for b in results
+                if (query.lower() in b['Title'].lower() or
+                    query.lower() in b['id'].lower())
+            ]
```
rodfersou: This way I can filter by part of the Title and by part of the ID.  I think the behaviour becomes better like this for the items without Title attribute.

src/collective/cover/static/contentchooser.js
```diff
@@ -50,20 +71,38 @@
         $(windowId + " ul.formTabs").tabs("div.panes > div");
     }
 
+    var parse_url = function(url) {
```
rodfersou: need this method to parse the plone default navigation url to get b_start attribute: "http://localhost:8080/Plone/teste/compose?b_start:int=4"

it just breaks the url, convert to int the value, and return a dict with values

src/collective/cover/static/contentchooser.js
```diff
@@ -171,216 +209,214 @@
...
 
 
-};
+    };
 
-coveractions.preInit();
+    coveractions.preInit();
 
 
-(function ($) {
```
rodfersou: Fixed jquery scope to be present to the entire script

src/collective/cover/browser/contentchooser.py
hvelarde: I think this is far from optimal; in what case you need to search by id?

src/collective/cover/static/contentchooser.js
hvelarde: me neither; give some clues

src/collective/cover/browser/contentchooser.py
rodfersou: If you test with collective.cover buildout there are some elements without title, and it becomes weird because you saw for example my-image there but when search for 'my-' didn't get the image

hvelarde: that's a bug in the tests that must be fixed

PR #503, opened 2015-04-13
Tidy custom-description-input by setting width rather than max-width

djowett:
Tidy "Edit Cycle2 Carousel Tile" view so description box lines up nicely with Title & URL box.


djowett:
Before:
custom-desc-before

After:
custom-desc-after

PR #505, opened 2015-04-13
Ship with more than one grid system

### LINKED ISSUE #504: Ship with more than one grid system
We need to ship cover with more than one grid system by default, so we will add 2 new 12-columns grid systems based on Bootstrap 2 and Bootstrap 3.

Bootstrap 2 (12 columns) will be the default one for new installations; for upgrades Deco (16 columns) will remain unchanged.

We need an upgrade step to update all layouts stored in the site.

rodfersou:
closes #504


hvelarde:
@cguardia please follow up this as it will affect your work on compatibility with Plone 5.

hvelarde:
@fredvd @mauritsvanrees please help me reviewing this

src/collective/cover/profiles/default/registry.xml
```diff
@@ -10,10 +10,10 @@
       <value_type type="plone.registry.field.TextLine"/>
```
rodfersou: * Removed class attribute
* Changed default values to 12 columns grid system

src/collective/cover/profiles/testfixture/registry.xml
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0"?>
```
rodfersou: this file is not needed (tests will use the cover default registry information)

src/collective/cover/layout.py
```diff
@@ -233,15 +233,12 @@ def update(self):
...
-
-    title = _(u'Deco (16 columns, default)')
-    ncolumns = 16
+class BaseGrid(object):
+    title = u''
+    ncolumns = 0
 
     row_class = 'row'
-    column_class = 'cell'
+    column_class = 'column'
```
rodfersou: if grid is bootstrap, the cell class was making the grid to style wrong, and it was breaking many RF tests, so for bootstrap grid systems we use a new class for columns

src/collective/cover/testing.py
```diff
@@ -154,27 +138,11 @@ def setUpPloneSite(self, portal):
 FIXTURE = Fixture()
 
 
-class MultipleGridsFixture(Fixture):
```
rodfersou: There is no need of this fixture as this will be the default behaviour.
I merged the tests from this fixture to the Integration tests

src/collective/cover/upgrades/__init__.py
```diff
@@ -222,3 +224,50 @@ def fix_persistentmap_to_dict(context):
...
+    settings.layouts = fixed_layouts
+    # Fix registry layouts
+
+    # Fix cover layouts
+    covers = context.portal_catalog(portal_type='collective.cover.content')
+    for cover in covers:
+        obj = cover.getObject()
+        logger.info('Remove css class from "{0}" cover layout'.format(obj.Title()))
+        obj.cover_layout = _remove_css_class_layout(obj.cover_layout)
+        obj.reindexObject()
```
hvelarde: I think there is no need to reindex the object as the ``cover_layout`` attribute is not used in the catalog

hvelarde:
@rodfersou we are missing a test for the upgrade step

src/collective/cover/config.py
```diff
@@ -25,7 +27,9 @@
     'News Item'
 ]
 
-DEFAULT_GRID_SYSTEM = 'deco16_grid'
+PLONE_VERSION = pkg_resources.require('Plone')[0].version
```
mauritsvanrees: This actually tries installing Plone, getting packages from the Internet if for some unimaginable reason Plone is not available yet. Sure, in practice that cannot happen, but it *does* try to do too much.  We should use this instead, which is a bit faster anyway:

    pkg_resources.get_distribution('Plone').version

See also the [plone.api docs](http://docs.plone.org/external/plone.api/docs/contribute/conventions.html#about-imports)

src/collective/cover/layout.py
```diff
@@ -257,8 +256,59 @@ def transform(self, layout):
...
+    """Bootstrap 3 grid system (12 columns)."""
+
+    grok.name('bootstrap3')
+    grok.implements(IGridSystem)
+
+    ncolumns = 12
+    title = _(u'Bootstrap 3')
+
+    def columns_formatter(self, columns):
+        prefix = 'col-md-'
```
mauritsvanrees: I wonder if we need and/or want to add `col-xs-` as well (and maybe `col-sm` and `col-lg`), with different values. But I guess that would need extra settings on the cover: `column-size-xs`, etcetera, including a way for the user to set that in the Plone UI on the layout tab. That sounds too complex, and certainly not something that would be needed in this pull request.

mauritsvanrees:
I looked at the code and tried it out. A few minor remarks above.

Good job! +1 from me.

djowett:
I'm glad to see a 12 column grid system by default, but can someone explain why the push towards bootstrap as a default, rather than a 12 col version of the *deco* grid system?  (Is Plone 5 moving away from deco gs?)

hvelarde:
@djowett Deco grid system is dead; AFAIK Bootstrap will be used in Plone 5; see:

https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/less/barceloneta.css

PR #507, opened 2015-04-20
Fixes performance issues when accessing the content tree

### LINKED ISSUE #494: Performance issues when accessing the content tree
When accessing the content tree tab on the content chooser we have noticed that the current script is far from optimal and sometimes just hangs:

selection_002

Things to be fixed:

* we should list all elements inside the folder
* folder contents should be shown with its original order, regardless if they are items or other folders (just like folder contents tab in Plone)
* when filtering we should wait for the user to stop typing before making a search (check Plone's [livesearch script](https://github.com/plone/Products.CMFPlone/blob/4.3.x/Products/CMFPlone/skins/plone_ecmascript/livesearch.js))
* filtering should be done only on the title index
* an spinner should be shown to let the user know there is a filtering in progress
* after filtering the order must continue unchanged
* we should show the item description in the title attribute
* we should use the ``grab`` cursor when selecting an item and ``grabbing`` when dragging it

hvelarde:
fixes #494


buildout.cfg
```diff
@@ -17,6 +17,7 @@ parts +=
     node
```
rodfersou: @hvelarde can I keep these configurations? helps me when debugging and creating RF tests

hvelarde: @rodfersou which ones?

src/collective/cover/browser/contentchooser.py
```diff
@@ -202,9 +203,18 @@ def jsonByType(self, rooted, document_base_url, searchtext):
         catalog_query['portal_type'] = self.filter_portal_types
         catalog_query['path'] = {'query': path, 'depth': 1}
         if searchtext:
-            catalog_query = {'SearchableText': '{0}*'.format(searchtext)}
+            catalog_query['Title'] = '{0}*'.format(searchtext)
+
+        brains = catalog(**catalog_query)
+        page = int(page, 10)
```
hvelarde: @rodfersou shouldn't be 20 here?

Issue #331, opened 2013-10-25
After installing c.cover, AJAX responses gets wrapped


rnixx:
After Installing collective.cover, AJAX responses gets wrapped by html / body when performing $.ajax calls.

Response before Install
![before_c cover](https://f.cloud.github.com/assets/246021/1408026/2838551c-3d7a-11e3-90e6-9d76829a1552.png)

Response after Install
![after_c cover](https://f.cloud.github.com/assets/246021/1408030/326b16b4-3d7a-11e3-9874-b46c6fab994c.png)

This breaks scripts depending on returned structure.

IMO it's not a good idea to overwrite generic API's at all.

petri:
It looks like a fix for plone.app.blocks was merged by @datakurre and subsequently released in p.a.b 2.0.0? Can anyone confirm whether that fixes this issue or are some changes still required here?

PR #516, opened 2015-04-24
Upgrade to plone.app.blocks 2.x [#331]

datakurre:
This pull is to see, what Travis thinks about this and discuss about this.


datakurre:
FYI: I haven't tried upgrading from older c.cover to a c.cover with p.a.blocks 2.0.0, but I don't recall any required changes in p.a.blocks configuration.

datakurre:
@hvelarde Build on Plone 4.2 fails, because p.a.blocks requires Plone 4.3. You added that requirement. Do you remember, what it p.a.blocks required Plone 4.3?

datakurre:
Checked. The layout behavior uses grok-free Dexterity 2.x API for defining the layout fieldset and Dexterity 2.x versions are default only from Plone 4.3 onwards.

Your call. Cover does not need must of it, so it could be refactored with conditional imports to be backwards compatible, if you wish so.

hvelarde:
let's make one more release and then we can drop support for Plone 4.2 in 1.0b2.

do you agree?

hvelarde:
@datakurre I made a rebase; could you please double check to see if I didn't broke anything?

datakurre:
@hvelarde Look OK and tests seem to pass.

hvelarde:
@datakurre what's the status on the new versions of plone.app.tiles and plone.tiles? should we upgrade those also?

datakurre:
@hvelarde Maybe one step at time? plone.app.tiles 1.x -> 2.x includes backwards incompatible tile deletion API change (tile deletion will be through similar URL and form as edit to allow some complex customizations and make it easy to support Plone 5's POST authorization tokens).

djowett:
+1 to "one more release"

djowett:
Can I ask what this upgrade gives to collective.cover?  I'm not questioning doing it, I just want to understand the benefits / impact

datakurre:
It fixes some side effects of c.cover 1.x by enabling the blocks transform only on c.cover views.

Daniel Jowett kirjoitti pe toukokuuta 22 17:47:26 2015 GMT+0300:

hvelarde:
@djowett it fixes #331

hvelarde:
thanks, @datakurre; you rock!

PR #513, opened 2015-04-21
Issue 284 - Add custom classes for row and columns

rodfersou:

PR #517, opened 2015-04-25
Filter CSS class vocabularies for column / row class choosers introduced in #513

djowett:
This small change allows (for instance) people to override classes shown in the drop down list for Columns & Rows configurations.  This could be used to (for example) filter the classes from TileStylesVocabulary.

The following demonstrates a simple override for Row & Column styles once this is in place:

In **overrides.zcml**:

```
<utility
    factory=".vocabularies.RowStylesVocabulary"
    provides="zope.schema.interfaces.IVocabularyFactory"
    name="collective.cover.RowColumnStyles"
    />
```

**vocabularies.py**

```
from zope.schema.vocabulary import SimpleTerm
from zope.schema.vocabulary import SimpleVocabulary

class RowStylesVocabulary(object):

    def __call__(self, context):
        items = []
        items.insert(0, SimpleTerm(value=u'row-default', title='Row Default'))
        items.append(SimpleTerm(value=u'row-other', title='Row Other'))

        return SimpleVocabulary(items)
```

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -257,3 +257,38 @@ If you switch from the default 16-column Deco grid to another grid with a differ
 these saved layouts will still contain a 16-column width and this can mock up your design in small ways.
 In that case,
 make sure you clear the default cover layouts and/or save your own layout with the correct number of columns.
+
+Row and Column Classes
+++++++++++++++++++++++
+
+``collective.cover`` enables you to set CSS classes on rows and columns in its layouts.
+These classes are those listed in the Styles field of the @@cover-settings view, which is
+also used for Tile styles.
+This does mean that rows, columns and tiles will have the same list of classes available for styling.
+If this is not satisfactory then the vocabularies ``collective.cover.RowColumnStyles`` and/or
+``collective.cover.TileStyles`` can be overridden in your own product as shown below:
+
+In **overrides.zcml**:
+
+    <utility
+        factory=".vocabularies.RowColumnStylesVocabulary"
+        provides="zope.schema.interfaces.IVocabularyFactory"
+        name="collective.cover.RowColumnStyles"
+        />
+
+**vocabularies.py**
+
+    from zope.schema.vocabulary import SimpleTerm
+    from zope.schema.vocabulary import SimpleVocabulary
+
+    class RowColumnStylesVocabulary(object):
+
+        def __call__(self, context):
+            items = []
+            items.append(SimpleTerm(value=u'row-default', title='Row Default'))
+            items.append(SimpleTerm(value=u'row-alternative', title='Row Alternative'))
+            items.append(SimpleTerm(value=u'column-default', title='Column Default'))
+            items.append(SimpleTerm(value=u'column-alternative', title='Column Alternative'))
+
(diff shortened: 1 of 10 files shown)


djowett:
fix: we *should* look up vocab by name in cover.py
tweak: multiple vocabs

hvelarde:
@djowett is my understanding that this PR will add the following tweak in case a vocabulary is overridden:

* if a user names a class with something starting with ``row-`` that class will be filtered from the tile and column class drop-down listing
* if a user names a class with something starting with ``column-`` that class will be filtered from the row and tile class drop-down listing
* if a user names a class with something starting with ``tile-`` that class will be filtered from the row and column class drop-down listing
* if a class does not starts with any of those strings then it should be listed everywhere

that must be documented somewhere.

is that fine? is that finished?

djowett:
@hvelarde no it doesn't do that (it's only a 3 line change after all), but it certainly should allow you to do that sort of thing pretty easily in your own product should you want to... No monkey patching required.

I was going along the lines of keeping complexity out of the product. But in the plone way, making simplicity overrideable. Does that sound good ?

However there's no way for it to distinguish between rows and columns yet, so not complete in that sense.


---- Héctor Velarde wrote ----

hvelarde:
ok, seems fair to me; do you want to document this on the developers guide?

djowett:
@hvelarde how about that?

djowett:
@hvelarde please see latest commit for implementation of what you mentioned above:


* if a user names a class with something starting with ``row-`` that class will be filtered from the tile and column class drop-down listing
* if a user names a class with something starting with ``column-`` that class will be filtered from the row and tile class drop-down listing
* if a user names a class with something starting with ``tile-`` that class will be filtered from the row and column class drop-down listing
* if a class does not starts with any of those strings then it should be listed everywhere

Still needs work on:

* Docs 
* adding some column & row classes to default profile (can you suggest any?)
* tests tweaking for new col & row classes
* test filtering (though I've tested manually)

djowett:
@fredvd @mauritsvanrees we need another opinion or two here.... would either of you have a chance to review this feature? 
Any questions pls ask

mauritsvanrees:
I have not tried it, I only browsed the code of this pull request. The idea seems useful and the code seems okay.

hvelarde:
yes, the idea seems useful but this is no longer a small change as proposed at the beginning; I'm still a little bit reluctant on merging it because I want to remove complexity from the package and no add more.

I think there has to be an easier way to do this without messing around JS and templates.

djowett:
Hi @hvelarde,  a few points to explain why this has grown from the small change:

1. This includes a fix for (the already merged) #513 - to title class selectors for Column & Row differently (that's why cover.py, the JS & templates were touched)
2. You asked for documentation to explain overriding the vocabularies, in the end it seemed easier to actually build the different vocabularies than to talk about how to do it!  I therefore plan to delete the documentation I added.  The elaborated formHelp should be sufficient.

This is still WIP, shouldn't take long to finish it, if it's worth it...

djowett:
closing as way out of date.

PR #519, opened 2015-05-21
Fixes RF carousel test

djowett:
The test was failing because the div with class 'textline-sortable-element' actually had a trailing space in the class attribute (and sometimes extra classes), and the test was too fragile to cope with that.


hvelarde:
awesome! thank you, very much!

PR #524, opened 2015-05-27
Remove carousel tile for core

### LINKED ISSUE #485: Should we remove carousel tile from the package core?
Lately we have had some discussions about many things on the carousel tile and some pull requests have been proposed to offer an alternative solution

* Replacing Galleria with Cycle2 (#442)
* Displaying a collection in a Bootstrap carousel (#470)

There are many solutions for carousels and everybody seems to have their own opinion on which one is the best.

So… should we just spin off the code of the carousel tile out of the core of this package to make it easier to maintain and leave people choose whatever JS plug-in they want to use?

hvelarde:
closes #485 

WIP, do not merge


djowett:
Good progress.... we need to be careful around UUIDSFieldDataConverter - which both covertile.* products need.  Should it stay in c.cover?

hvelarde:
thanks, @djowett, it's gonna be more difficult that I was expecting; I'll continue this afternoon.

djowett:
@hvelarde this never got removed eh? Which carousel are you using now? (I'm not bothered if things work)

hvelarde:
@djowett we're still using Galleria and yes, it works; as that's a breaking change we probably want to replace it with Cycle2 or maybe the Bootstrap one as suggested by @ebrehault.

PR #533, opened 2015-06-23
Fix StaleElementReferenceException in some cases.

idgserpro:
When trying to use the keyword "Compose Cover" in some packages that
use collective.cover as a dependency, the error

    StaleElementReferenceException: Message: Element not found in the cache - perhaps the page has changed since it was looked up

happens in an intermitent manner. Adding

    Wait Until Keyword Succeeds  5 sec  1 sec

avoids this problem.

This solution was already discussed with the community in

http://www.coactivate.org/projects/barcelona-sprint/lists/barcelona-sprint-discussion/archive/2013/02/1360419825199/forum_view


hvelarde:
thanks!

PR #536, opened 2015-07-15
Add new permission to restrict code inclussion on the Embed tile

### LINKED ISSUE #297: Content of Embedding Tile is not security checked
It is possible to add arbitrary HTML / JavaScript code into the Embedding Tile. 

```
<script type="text/javascript">
<!--
alert("Something evil!");
//-->
</script>
````

The safe filter of Plone should be applied to some extent.

hvelarde:
Now, by default, only Managers and Site Administrators are able to insert code in the tile.

closes #297


hvelarde:
@davisagli any comments are welcome!

PR #537, opened 2015-07-17
Update versions of dependencies

hvelarde:
I need help to test this and verify we are not breaking anything.

@rodfersou @fredvd @djowett can you help?


hvelarde:
@datakurre tests seem to be passing; is this safe enough?

datakurre:
@hvelarde Unpinning p.tiles should be safe. Unpinning p.a.drafts should also be safe, because its Archetypes-support is now disabled by default, and for DX it shouldn't do anything until a Draftable-behavior is enabled. (Unpinning p.a.tiles would not be safe, because the changed delete-tile API. p.a.blocks, on the other hand, depends on CMFPlone 4.3, but only, because it's using DX 2.x API for its behavior. If that's blocker for c.cover, maybe p.a.blocks should be fixed to be backwards compatible with e.g. conditional imports.)

fredvd:
@hvelarde We're not using plone.app.drafts, but I upgraded both to the latest versions locally (plone.app.drafts 1.0b3 and plone.tiles 1.4) and editing/changing content seems to work fine so far.

hvelarde:
thanks, guys!

PR #538, opened 2015-08-14
Fix caroussel edit compose on IE11

fredvd:
Fix textlinessortable widget for IE11 where IE11 mangles multiform POST data. This fixes removing items from the Caroussel compose widget in IE11.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.0a12 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


fredvd:
@hvelarde Does this mean with the second check that CI now passes?

hvelarde:
seems to me a test is failing; could you please take a look on it? also, please squash all commits into one and ping me if you need to restart a test job

fredvd:
@hvelarde squashed. 

Test failing was a random one. Somehow you restarted it as well (allthough I can also restart myself on travis-ci.org), when I checked the page here all checks were green already before I changed anything.

PR #546, opened 2015-09-09
Add Cache-Control headers to tile configuration and edit overlay views

fredvd:
This prevents Internet Explorer from caching the XHR GET requests for these views. What would happen is that you would see the previious (old) field info if you configured or edited a tile, saved and re-opened the same dialog in IE.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ There's a frood who really knows where his towel is.
 1.0a12 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
+- Add Cache-Control headers to the @@configure-tile and @@edit-tile views to prevent Internet Explorer from caching the XHR GET requests for these views. What would happen is that you would see the previious (old) field info if you configured or edited a tile, saved and re-opened the same dialog in IE.
+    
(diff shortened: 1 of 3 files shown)


hvelarde:
awesome! thanks a lot

fredvd:
@datakurre Could this also be an issue in plone.app.mosaic where the underlying tile infrastructure is used, or is mosaic only using post requests to fetch tile data/configuratin/rendering? I have no experience with p.a.mosaic's code.

datakurre:
@fredvd Good to know. Sounds possible, but could probably be fixed also in javascript side by adding e.g. timestamp into request.

hvelarde:
@datakurre let us know if you find a better way to implement this.

datakurre:
@hvelarde I have not yet tested, if this happens in Mosaic with IE. At least one difference is that Mosaic uses the plone mockup modals, and there might be something for this already in mockup modal.

fredvd:
@hvelarde I'm trying to fix some stuff with covertile.cycle2 and I just noticed the xhr requests for @@updatetilecontent are also done using GET requests without Cache-Control headers. My gut feeling is that dropping extra items on list & carousel tiles in IE11 will show the same stale content isuses.

hvelarde:
@fredvd good to know; FYI @rodfersou

rodfersou:
:+1:

PR #551, opened 2015-09-17
All tiles made searchable

lewicki:
New ISearchableText interface to adapt on tiles to make them searchable. All custom tiles may provide their own adapters.


hvelarde:
seems we have some tests failing; could you please take a look on that?

also, don't forget to add a line into the changelog and squash all commits into one after you finish.

PR #552, opened 2015-09-23
Use specified scale on tile to fetch scaled image, instead of full sized

alecpm:
When cloning an image from a content object, fetch the scale configured on the tile to ensure we get a correct crop from plone.app.imagecropping and don't waste space storing a full sized image that won't be rendered.


hvelarde:
awesome!

PR #553, opened 2015-09-25
Reimplement PR #409

alecpm:
See:

https://github.com/collective/collective.cover/pull/409

This got undone in a recent commit.  This patch adds content path to title tooltip, to ensure content listed in search popup is completely unambiguous.


hvelarde:
@alecpm tests are failing because an unused import; could youplease fix that and squash your commits?

Issue #549, opened 2015-09-12
Why don't generate UID from javascript?


rodfersou:
Today we have [a view to generate UID](https://github.com/collective/collective.cover/blob/master/src/collective/cover/layout.py#L200-L214) for tiles.

There is [a long stack overflow](http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript) with many option on how to generate UID from javascript.

Also, this stack overflow post point to [some implementations like this one](https://github.com/LiosK/UUID.js) that are RFC-compliant.

hvelarde:
and how you will generate programmatically tiles for sites like [IDGB](https://github.com/plonegovbr/brasil.gov.portal)?

rodfersou:
@hvelarde I didn't get your point.. I just want to know why do I need this view since I can generate the uid with javascript.

hvelarde:
if you think is going to work then create a branch with the changes and open a pull request so we can evaluate it.

PR #554, opened 2015-09-29
Generate the tiles UUID in Javascript

### LINKED ISSUE #526: Same ID for tiles on IE10/11
When adding two new tile to a column (with boostrap3 grid) they will have the same ID. This results in storing the data under the same tile and getting an exact copy of each other. This only happens in IE10/11. 

screen shot 2015-06-03 at 14 45 23

See attached screenshot of resulting HTML.

frapell:
Right now, when a tile is dropped into a column from the layout view, an ajax call happens to the server, to the @@uid_getter view.

This is a problem, not only because of the overhead of an additional unnecessary request, but also when a site is behind varnish, it can happen that the result of @@uid_getter is responded by varnish with an old, already used uuid, having then 2 or more tiles with the same uuid, provoking that changes to one tile affect the other(s)

This PR adds a Javascript function that will generate the uuid, avoiding having to do a request to the server.

fixes #526


frapell:
@hvelarde when you have a chance, please review this. Thanks!

frapell:
Umm... tests pass in my local instance... not sure what's up with Travis...

hvelarde:
@frapell please remove the redundant code in Python

@rodfersou this is related to #549

frapell:
@hvelarde done

frapell:
@hvelarde How can I tell Travis to ignore a test or not fail at least? I don't want to remove test_layout completely, eventually there should be unit tests in there testing all different methods for the layout, but is out of the scope of this PR.

FWIW, Travis ends with
Total: 249 tests, 0 failures, 0 errors in 16 minutes 22.163 seconds.

but it fails with message TypeError: Module collective.cover.tests.test_layout does not define any tests and that is because I of course removed the test for the uid_getter

hvelarde:
@frapell you can freely remove that file if it has no more tests inside; I expect to have your JavaScript code tested before we can merge.

hvelarde:
@frapell I made a little research and seems you're not generating a real UUID; take a look at this Stack Overflow answer on RFC4122 UUIDs: http://stackoverflow.com/a/2117523

I would suggest also to make a check on Python code to avoid people messing around it.

thinking a little bit I don't know if we really want to do this; the caching issues could be solved more easily by adding some HTTP header.

@fredvd opinions?

frapell:
@hvelarde we don't really *NEED* a UUID... it just needs to be a random string that doesn't repeat between tiles for the given context... even so, assuming you don't have more than 25 tiles per cover, you could just use letters from a to z, so this is implementation is more than enough.

Also, we cannot do unit test of Javascript, until we can have a refactoring to move the code to be mockup-compatible. Having all the funcitonal tests passing, means the id for the tiles are being generated correctly

frapell:
BTW, this is not only a cache issue, is a way to avoid doing unneeded requests to the server when you can do it from JS

rodfersou:
@frapell I think we should use this implementation of uid generation since it uses a timestamp:
```js
function generateUUID(){
    var d = new Date().getTime();
    var uuid = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
        var r = (d + Math.random()*16)%16 | 0;
        d = Math.floor(d/16);
        return (c=='x' ? r : (r&0x3|0x8)).toString(16);
    });
    return uuid;
}
```
As you said before, it is not a perfect solution, a random strind would work fine, but this way we ensure that the same browser would not create 2 equal timestamps.

frapell:
I don't mind changing it... I would prefer my approach because it is easier to understand... I look at the proposed function and I don't really understand what is going on... I would be REALLY amazed to see 4 calls to Math.random() appended together to generate the same string twice in ~10 elements...

rodfersou:
@frapell just set the first 13 bytes with the Date().getTime() result in your approach, and you will get the same result + easy reading for you.

Since I have some background with electronics I don't mind in read the method I pasted before, but you are right, it is better more people can read.

Or maybe we can add a new dependency to this package https://github.com/crypto-utils/uid-safe

rodfersou:
@frapell sorry.. now I saw that it mix the timestamp numbers with the generated char:

(generated_timestamp_char + generated_char) % 16

rodfersou:
The link I posted is for node.. this could be a javascript option https://github.com/pnegri/uuid-js (there are many other implementations)

rodfersou:
Other option https://github.com/broofa/node-uuid

hvelarde:
again, guys, just 2 lines in Python and we get rid of this issue:

```python
self.request.RESPONSE.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate')
self.request.RESPONSE.setHeader('Expires', '0')
```

and best of all, is already tested...

frapell:
Again, I don't mind if someone changes the function to another implementation if they want to... I'm sorry but I don't have the time to do a refactor of a working implementation just for the sake of changing it...

As I explained, this doesn't need to be following any standard, or including any extra dependency, and the provided implementation will work in 99.99999999% of the cases

hvelarde:
@frapell I already give a solution above, easier and cleaner; test it with your use case and let me know.

davilima6:
+1 on merging this PR. Dependencies should be avoided whenever possible and we can easily save network traffic and make UX smoother with @frapell's solution. If any of you think it's worrisome maybe we could add a comment for integrators not to confuse this implementation with Plone's uuid machinery, it's just similar concept, right?

frapell:
@davilima6 that is correct... just a random string

frapell:
@davilima6 you got me thinking, UX could be even smoother if we shorten this to just the first while loop, giving an 8 characters length string, which should be more than enough to be unique among tiles in the same cover (heck... it's even enough to reference github commits ;) )

hvelarde:
Plone 4.3 and 5.0 pin jQuery UI 1.10 and we already have the following function there:

http://api.jqueryui.com/uniqueId/

we are going t deprecate Plone 4.2 in next version anyway.

hvelarde:
@rodfersou told me that the implementation of uniqueId is not good enough.

@frapell to make it easier and cleaner use just a timestamp (adding a letter at the beginning); I think something like `Math.floor(Date.now() / 1000)` will do the job.

I already make a rebase; do a squash when you finsih.

fredvd:
I observed bedore that the uidgetter also was called but in my case locally it did have a cache-control, which was maybe because it's still a grok'ed view or something. 

But not to flog a dead horse, just do it client side with a random function that is random enough. Never reallised that tile registration is always done on context. So there's no change that we'll ever ever want to or could refer to a tile instance 'globally' as without also passing in context? (Didn't find it so far). 

To absolutely be sure that you have no id conflict you could add an extra property 'new' on first submission and check server side that no with this id has been created before on the context. That would throw a red flag and show any future caching issues.

frapell:
@fredvd Not entirely sure what you mean... tiles are not *really* registered anywhere, all it takes is to render a <div> with a data-tile attribute, which holds something like "@@collective.cover.richtext/random_string" (tiles id are saved in the cover itself, not relevant here...)

this pulls the tile data from an annotation in the current context using the "random_string" as key.

As soon as you change the "random_string", data in the tile will change immediately, this is what triggered a bug for us, where the ajax call to the @@uid_getter will be replied by varnish and 2+ tiles would be showing the same data.

Since this is context aware, and I cannot imagine a cover having more than 30 tiles, having a randomly generated string with 32 hex characters should work fine 99.999999999999% of the time, and checking if there is already a tile with the same uid shouldn't be needed really.

The provided implementation works fine, fixes the problems, it is tested (not automated, but functional tests still pass 100% of the time) and ready to be merged.

hvelarde:
@frapell this inot ready to be merged: you're missing a rebase and a squash, just to begin.

davilima6:
+1 for merging after squash. New and better implementations can always be requested in other PR so my 2c is release early, release often.

hvelarde:
@davilima6 sometimes releasing early and often is not a good idea, specially on packages as complex as this one; I prefer to spend a lot of time now discussing before merging than spend 5x a lot of time later fixing half baked implementations as has happened before.

davilima6:
I wouldn't suggest merging twice if this seemed half baked to me but I guess you already made your mind.

hvelarde:
yes, is open source, but if something ends broken **we** will be blamed and, as many productions sites depend on cover, we will have to dedicate our time (unpaid) fixing problems we didn't create; that's my point.

in this specific case, for me is OK to move the functionality from Python to JS, but I'm not sure if I want to change the implementation from UUID v4 to something else. to release often and early you have to move one thing at a time, not many.

hvelarde:
@rodfersou tests are failing for jQuery 1.8 and 1.9; we need to fix that before merging.

hvelarde:
the failures seem not to be related with these changes; I'm going to open a new issue to deal with that.

PR #556, opened 2015-10-20
All tiles made searchable

lewicki:
New ISearchableText interface to adapt on tiles to make them searchable. All custom tiles may provide their own adapters.
Revisited.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.0a12 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
+- All tiles made searchable.
+  [lewicki]
+
 - Use "application/javascript" media type instead of the obsolete "text/javascript".
   [hvelarde]
 
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -196,6 +196,29 @@ supported.
 .. _`Using tiles to provide more flexible Plone layouts`: http://glicksoftware.com/blog/using-tiles-to-provide-more-flexible-plone-layouts
 
 
+Searchable text in new tiles
+++++++++++++++++++++++++++++
+
+The content of new tiles is not searchable by default. To enable that functionality
+you need to register a new adapter for the ISearchableText interface. It is as simple
+as copying the code provided for ``plone.tiles.BasicTile`` and adding a new ZCML 
+registration.
+
+.. code-block:: python
+
+  class SearchableBasicTile(object):
+      implements(ISearchableText)
+
+      def __init__(self, context):
+          self.context = context
+
+      def SearchableText(self):
+          context = self.context
+          return u'{0} {1}'.format(context.data['title'] or '', 
+                                   context.data['description'] or '')
+
+
+
 Grid Systems
 ^^^^^^^^^^^^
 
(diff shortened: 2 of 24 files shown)


hvelarde:
@rodfersou could you please help me reviewing this?

@mauritsvanrees @fredvd @djowett opinions?

fredvd:
The option to have multiple tile typees indexable and that you can add indexing for custom (content) tiles is great !!

What concerns me is that for navigation/indexing there are use cases where you don't want the covers to be indexed (fully) because they serve as landing pages and navigation to real content pages on one subject. Covers are most of the time aggregates. When a visitor searches for "my specific page title" you want them to find the content page in the search results, not the 5 covers on which this title is also mentionned.

It's a minor point because this pull request doesn't do the above because only title, desciption and some other general info of 'listy' tiles' is inserted in to the default IsearchableText, not the list items themselves, but adding titles of a list tile to the searchable text for the cover is imho of limited use and might also pollute search results as these are mostly headers like "relevant news", "the latest events" etc. Thing here is that we probably don't want to change behaviour of installed base. 

@lewicki What's your use case for adding title and description of list/collection tiles to the covers searchabletext? Is it a real user demand, or is the ISearchableText mainly there to provide indexing for custom 'content' tiles? then I'd say don't provide adapters for the listy tiles at all or comment them out, then integrators can add them if deemed necessary.

lewicki:
@fredvd There is no usecase. I needed searching functionality for custom tiles so I implemented it in adaptable form. I just provided the example implementation for basic set of tiles. I may revert that part.
Anyone else to comment?

mauritsvanrees:
Seems okay.

djowett:
Hi, FWIW I'm fairly aligned with @fredvd here.  I think it's a good feature to have the option, but the default behavior should be sensible.  It is important that the first search result on a search matching a document / file title should be the document / file itself NOT a cover which has a tile referencing the object.

I think the RichText and Embed tiles should remain & the implementations for other tiles be reverted.
@lewicki can you say what your custom tiles did?

hvelarde:
@lewicki do you mind adding a section on the developers documentation? this seems useful to me.

lewicki:
@djowett Nothing special. Something between basic tile and RichText. Specific customer request.
@hvelarde I'll update documentation as well.
I'll revert some adapters then.

djowett:
@lewicki so similarity to Richtext tile justifies making your custom tile searchable.  Did you make any other tiles searchable for that customer?

fredvd:
@lewicki The decoupling of indexing and tiles is great, until now only richtext was indexed at all and no pluggability. 

Just comment out the registration for other tiles than Rich Text and Basic or remove them except for one that is a commented example for further customisation for other cover integrators. Thanks for the nice addition!

hvelarde:
yes, I agree; after adding the documentation it would be very easy for integrators to activate this feature for more tiles. the documentation can, for instance, take the Basic tile and make it searchable.

to keep backwards compatibility I think we should left only the RichText tile as searchable; let me know if you all agree on that.

djowett:
@hvelarde I'm happy with only the RichText tile searchable.
I'd also be happy with making Embed tile searchable as well, if others thought that sensible.

lewicki:
I updated the code and added some info to developers documentation. I left the adapter for BasicTile, EmbedTile and RichtextTile. Regarding backward compatibility - I would consider not having BasicTile being indexed as a bug not feature.

mauritsvanrees:
Quick remark. I see that Travis fails because of a version conflict for coverage. This is probably caused by a new createcoverage version that is pulled in which now requires a newer coverage. See https://github.com/reinout/createcoverage/pull/1

hvelarde:
@mauritsvanrees yes, I already fixed that on master, but now we have some other tests failing with newer versions of jQuery.

PR #560, opened 2015-11-04
Issue 559 - Review alt attribute of images

rodfersou:


src/collective/cover/tiles/templates/tile_macros.pt
```diff
@@ -18,22 +18,3 @@
   </span>
```
rodfersou: This macro is not used AFAIK.

hvelarde: this has to be documented on the changelog.

hvelarde:
@polyester we need your expertise here: on `alt` attributes, should we use the Title or the Description of the items?

polyester:
Can never be completely guaranteed; the rules say you have to provide a 'meaningful' `alt` tag. 

For mechanical validators, anything will do, obviously, but it seems @rodfersou is making the right choice here *(as I read the code...)* : use `description` when it is there (because then it will have semantic content), use `title` if it isn't there because at least that will be filled in (even with just the DMC2948274893.jpg of the image if you're not careful...)

hvelarde:
@polyester understood; @rodfersou I think it would be great if we create a function called alt in all tiles; that way we can put the logic inside and maybe, in the future, we can provide a way to modify that.

rodfersou:
@polyester thank you for your advise.
@hvelarde nice idea, the main problem is when a tile have more than one image (ex.: tile collection or list) let's think more about it! :wink:

PR #572, opened 2015-12-14
Reindent code

### LINKED ISSUE #487: Implement drag and drop among tiles
We have a requirement from one of our customers to implement this feature, as they spend endless hours editing the front page of their site just to move items among tiles.

Currently we only support to drag and drop content inside a list tile to change the order of an item on it. We need to be able of moving items as well.

As implementing this in all tiles will be very complex, I suggest to start supporting only basic and list tiles (that's probably 80% of the use cases):

* moving an item from a basic tile to another tile: all customized data must be copied; the first tile must be emptied; on a full list the effect must be the same as adding a new item on the list: the last element is gone (we need to check the current behavior)
* moving an item from a list tile to another tile: all customized data must be copied; the item is removed from the list in the first tile

## Risks ##
* We can lose information (i.e. a customized image) on custom fields not supported in the list tile (we need to check this)
* To avoid losing information in case of an error, is better to copy the information in the second tile before erasing it from the first one
* I haven't tested if the functionality included in #451 has some unexpected consequences on the current feature
* We need to update the RF tests and that could be painful

rodfersou:


hvelarde:
@rodfersou please don't; do the indentation stuff outside this PR

PR #573, opened 2015-12-15
Issue 487 - Implement drag and drop among tiles

### LINKED ISSUE #487: Implement drag and drop among tiles

rodfersou:
closes #487


src/collective/cover/layout.py
```diff
@@ -142,6 +144,25 @@ def accepted_ct_for_tile(self, tile_type):
...
+        data_mgr = ITileDataManager(tile)
+        data = data_mgr.get()
+        uid = data.get('uuid', None)
+        if uid is None:
+            return
+        catalog = api.portal.get_tool('portal_catalog')
+        results = catalog(UID=uid)
+        if len(results) == 0:
+            return
+        return results[0].portal_type
```
rodfersou: ** replace this code to uuidToObject method

src/collective/cover/tiles/list.py
```diff
@@ -222,8 +223,13 @@ def populate_with_uids(self, uuids):
         data_mgr.set(old_data)
         notify(ObjectModifiedEvent(self))
 
-    # XXX: This should be renamed to replace_with_uuids
     def replace_with_objects(self, uuids):
+        warn('Method replace_with_objects() is deprecated. Use ' +
+             'with replace_with_uuids() method instead.',
+             DeprecationWarning, stacklevel=2)
+        self.replace_with_uuids(uuids)
```
rodfersou: use this example of deprecation https://github.com/plone/plone.app.layout/blob/master/plone/app/layout/viewlets/common.py#L46-L47

src/collective/cover/tests/test_drag_and_drop_among_tiles.robot
```diff
@@ -0,0 +1,55 @@
...
+${carousel_tile_location}  'collective.cover.carousel'
+${image_selector}  .ui-draggable .contenttype-image
+${basic_tile_selector}  [data-tile-type=collective\\.cover\\.basic]
+${banner_tile_selector}  [data-tile-type=collective\\.cover\\.banner]
+${list_tile_selector}  [data-tile-type=collective\\.cover\\.list]
+
+*** Test cases ***
+
+Test Drag And Drop Among Tiles
+    [Tags]  Expected Failure
```
rodfersou: I don't know what is going on here.. locally it works..

rodfersou:
@hvelarde how about a christmas release with this pull request? :smile_cat:

src/collective/cover/tiles/collection.py
```diff
@@ -77,6 +77,7 @@ class ICollectionTile(IPersistentCoverTile):
 
     uuid = schema.TextLine(
         title=_(u'UUID'),
+        required=False,
```
hvelarde: why?

src/collective/cover/tiles/contentbody.py
```diff
@@ -13,7 +13,11 @@
 
 class IContentBodyTile(IPersistentCoverTile):
 
-    uuid = schema.TextLine(title=u'Collection uuid', readonly=True)
+    uuid = schema.TextLine(
+        title=_(u'UUID'),
+        required=False,
```
hvelarde: why?

src/collective/cover/tiles/pfg.py
```diff
@@ -25,8 +24,11 @@ class IPFGTile(IPersistentCoverTile):
         required=False,
     )
 
-    form.omitted('uuid')
-    uuid = schema.TextLine(title=u'Collection uuid', readonly=True)
+    uuid = schema.TextLine(
+        title=_(u'UUID'),
+        required=False,
```
hvelarde: why?

src/collective/cover/browser/cover.py
```diff
@@ -219,6 +221,56 @@ def render(self):
             raise BadRequest('Invalid parameters')
 
 
+class MoveTileContent(grok.View):
```
hvelarde: I think is time to start getting rid of Grok on new views

src/collective/cover/browser/cover.py
```diff
@@ -219,6 +221,56 @@ def render(self):
...
+        else:
+            origin_tile.delete()
+
+    def render(self):
+        """Render a tile after populating it with an object."""
+        origin_type = self.request.form.get('origin-type')
+        origin_id = self.request.form.get('origin-id')
+        target_type = self.request.form.get('target-type')
+        target_id = self.request.form.get('target-id')
+        if not all([origin_type, origin_id, target_type, target_id]):
```
hvelarde: nice! I didn't knew about `all()` and `any()`

src/collective/cover/browser/cover.py
```diff
@@ -219,6 +221,56 @@ def render(self):
...
+
+    grok.context(ICover)
+    grok.require('cmf.ModifyPortalContent')
+
+    def _move_all_content(self, origin_tile, target_tile):
+        """Move all content from one tile to another tile"""
+        origin_dmgr = ITileDataManager(origin_tile)
+        origin_data = origin_dmgr.get()
+        if IListTile.providedBy(origin_tile):
+            uids = origin_data.get('uuids', [])
```
hvelarde: why are you using `uid` instead of `uuid`?

src/collective/cover/browser/cover.py
```diff
@@ -271,7 +323,7 @@ def render(self):
         uid = self.request.form.get('uid')
         if tile_type and tile_id and uid:
             tile = self.context.restrictedTraverse('{0}/{1}'.format(tile_type, tile_id))
-            if isinstance(tile, ListTile):
+            if IListTile.providedBy(tile):
```
hvelarde: I think this is wrong; are you sure a tile inheriting from list tile also provides `IListTile`?

src/collective/cover/layout.py
```diff
@@ -142,6 +144,22 @@ def accepted_ct_for_tile(self, tile_type):
 
         return json.dumps(accepted_ct)
 
+    def content_type_for_tile(self, tile_type, tile_id):
```
hvelarde: I know you don't like `get_` and `set_`, but is perfectly fine to use them and names get better: `get_content_portal_type` and `get_content_uuid` for instance

src/collective/cover/layout.py
```diff
@@ -142,6 +144,22 @@ def accepted_ct_for_tile(self, tile_type):
...
+        tile = self.context.restrictedTraverse('{0}/{1}'.format(str(tile_type), str(tile_id)))
+        data_mgr = ITileDataManager(tile)
+        data = data_mgr.get()
+        uid = data.get('uuid', None)
+        if uid is None:
+            return
+        obj = uuidToObject(uid)
+        return obj.portal_type
+
+    def content_uid_for_tile(self, tile_type, tile_id):
```
hvelarde: we're lacking documentation here

src/collective/cover/tiles/file.py
```diff
@@ -93,17 +93,18 @@ def is_empty(self):
...
-                          'description': description,
-                          'download': True,
-                          'uuid': uuid,
-                          })
+        if obj.portal_type not in self.accepted_ct():
+            return
+        title = safe_unicode(obj.Title())
+        description = safe_unicode(obj.Description())
+        data = {
+            'title': title,
```
hvelarde: move `safe_unicode(obj.Title())` to here; one less line

src/collective/cover/tiles/file.py
```diff
@@ -93,17 +93,18 @@ def is_empty(self):
...
+            'description': description,
```
hvelarde: move `safe_unicode(obj.Description())` to here; one line less

src/collective/cover/tiles/collection.py
rodfersou: just write the same way for all tiles

src/collective/cover/tiles/contentbody.py
rodfersou: just write the same way for all tiles

src/collective/cover/tiles/pfg.py
rodfersou: just write the same way for all tiles

src/collective/cover/browser/cover.py
rodfersou: ;)

src/collective/cover/browser/cover.py
rodfersou: I double check this and it works fine like this (it provides IListTile for tile List and tile Carousel)

rodfersou:
@hvelarde green! \o/

src/collective/cover/tiles/list.py
```diff
@@ -390,7 +395,7 @@ def __init__(self, context):
         self.context = context
 
     def getUIDs(self):
```
rodfersou: I can't change the name to getUUIDs because it is used into PFG

src/collective/cover/tiles/list.py
```diff
@@ -403,7 +408,7 @@ def __init__(self, context):
...
```
rodfersou: I can't change the name to getUUIDs because it is used into PFG

src/collective/cover/tiles/list.py
```diff
@@ -416,6 +421,6 @@ def __init__(self, context):
...
```
rodfersou: I can't change the name to getUUIDs because it is used into PFG

hvelarde:
@rodfersou a couple of issues:

* we're lacking end user documentation describing the new feature
* we're lacking an upgrade step to cook the resources
* can't move content out of a RichText tile
* got an error when trying to move content from one manually populated tile to another

having seen that, I think it could be important to check if the tile has an associated UUID **before** showing the cursor hand that indicates content can be moved from one tile to another

rodfersou:
@hvelarde I think all issues are ok now

docs/end-user.rst
```diff
@@ -192,6 +192,64 @@ steps:
...
 
+Move content between Tiles
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To manage a big site you will eventually need to change an important new from
+one place to another of your Cover.
+
+If you move the content data from one tile to another, it will behave just like
+if you deleted the content from the first tile and added the content data into
+the second tile (this feature works for every tile created after version 1.0a13).
```
hvelarde: @rodfersou why?

docs/end-user.rst
```diff
@@ -192,6 +192,64 @@ steps:
...
+   that Tile, the box will turn green. If not, it will remain gray.
+
+#. Release the item. Information from that content will be moved
+   to the chosen tile and will be visible on the cover.
+
+#. Repeat the process to move content to each one of the other Tiles.
+
+#. To visualize the final result, select the "View" tab.
+
+For older tiles, it will work just to move content for Tiles with same type.
```
hvelarde: @rodfersou why?

src/collective/cover/upgrades/configure.zcml
```diff
@@ -3,5 +3,6 @@
   <include package=".v11" />
   <include package=".v12" />
   <include package=".v13" />
+  <include package=".v14" />
```
hvelarde: @rodfersou this upgrade step is not needed; you already did that on `v13` and that's the one to be used on next release.

docs/end-user.rst
rodfersou: I didn't get your point.. where in this text we are talking about recreate the whole front page?!

src/collective/cover/upgrades/configure.zcml
rodfersou: ok, upgrade step removed

docs/end-user.rst
rodfersou: I changed the text from "every tile created" to "every content dropped"

hvelarde:
@rodfersou we still have this issue:

* add a news item with an image on it
* create a cover with 3 tiles: 2 basic and 1 list
* configure one basic tile to show the image, and the other to not show it
* drag & drop the news item into the list tile
* drag & drop the item into the basic tile that doesn't show the image
* drag & drop it now to the basic tile that shows the image

the image is lost.

rodfersou:
I forgot about this issue.. I think changing the order of the things will make this work..

rodfersou:
I added a new verification [here](https://github.com/collective/collective.cover/pull/573/files#diff-08f9d36638ec067fb66f20856d531aa3R255) to check if value is not none, then the invisible data will not change the content data.

rodfersou:
@hvelarde green \o/

PR #577, opened 2016-01-13
Remove hard dependency on plone.app.stagingbehavior as that package is no longer needed in Plone 5

hvelarde:
Under Plone < 5.0 you should now explicitly add it to the eggs part of your buildout configuration to avoid issues while upgrading.


hvelarde:
@alecpm @cewing @cguardia @mauritsvanrees could some of you make review on this?

src/collective/cover/setuphandlers.py
```diff
@@ -1,8 +1,21 @@
...
 
+import logging
+import pkg_resources
+
+try:
+    pkg_resources.get_distribution('plone.app.stagingbehavior')
+except pkg_resources.DistributionNotFound:
+    IStagingSupport = None
+else:
+    from plone.app.stagingbehavior.interfaces import IStagingSupport
```
rodfersou: isn't it better to get rid with the else clause and move this line of code (17) into the try clause (line 14) ?

IMHO it will becomes more readable.

hvelarde: usability 101: don't make me think; the `try` block contains only the line that could rise the exception; if I add 2 lines there, then I have to think which one causes the exception.

rodfersou: if you prefer a code more usable than readable, no problem for me :)

but I'll keep my way to think and code.. so prepare yourself to see different syntax on the `try` blocks I write in my PRs

mauritsvanrees: The code between a try/except should be as small as possible. So I like it how it currently is. I don't think it's less readable.

rodfersou: about your phrase `don't make me think`, I look different.. `else` clause for me looks like an alien.. always when I see it in a `try` block I need to remember what it means.. and when I remember what it means I mentally tie the try block with the else block and re-read everything.. so I need to think a lot more.

rodfersou: for me `else` in a `try` block just makes sense if I really want the exception to be raised, and treat it in another layer of the system.

but in this case I still don't use the `else` clause.. I try to arrange the lines I expect to raise the exception out of the `try` clause.

PR #579, opened 2016-01-14
Refactor link integrity feature on tiles

### LINKED ISSUE #578: Link integrity not working
Link integrity has been broken for some time now; we must re-implement it taking into account how this is going to work in Plone 5, where this feature was simplified to handle only deleting an item in the Plone UI.

According to first part of [PLIP 125](https://plone.org/products/plone/roadmap/125), the use cases could be something like this:

* user adds 2 images, A and B
* user adds a cover object with 2 tiles: one Basic and one RichText
* user drops image A on the Basic tile and references image B inside the RichText tile
* in Plone 4 those references must be saved on the `reference_catalog` in [Products.Archetypes](https://github.com/plone/Products.Archetypes/tree/1.9.x); on Plone 5 those references must be saved using [zc.relation](https://github.com/zopefoundation/zc.relation).
* user tries to delete image A and B; in both cases she gets a warning saying: "The image X is used in the page ABC, are you sure you want to delete it?"

Current implementation registers an event handler to take care of `IObjectClonedEvent` (I don't know why or what for), `IObjectAddedEvent` and `IObjectModifiedEvent`; it also registers a different handler for `IObjectRemovedEvent`.

I propose to create a RF test and then fix the feature to do only what is proposed above.

We should get rid also of the hard dependency on Products.Archetypes.

hvelarde:
fixes #578


hvelarde:
@idgserpro @winstonf88 this could have implications on brasil.gov.portal; please follow up.

PR #583, opened 2016-01-20
Issue 347 - Add RSS tile

rodfersou:


rodfersou:
@hvelarde should I register a new profile for the upgrade step?
I think it is safer if my upgrade step just add the tile to the tile list and don't change the other lines into ``registry.xml``

davilima6:
Nice one, thx!

rodfersou:
@davilima6 this is just a POC.. the real solution for collective.cover will be something different, but you can follow this implementation in your own site if you need it now.

Issue #586, opened 2016-02-18
Strange behavior of Plone Site when running bin/test


idgserpro:
We're trying to debug a test but the Plone Site from `bin/instance` is different from `bin/test` Plone Site.

How to test: run bin/instance in a bootstrapped collective.cover. Go to the ZMI and portalquickinstaller. You should see something similar to this image:

![selecao_124](https://cloud.githubusercontent.com/assets/8203476/13155865/f984bef8-d666-11e5-9d78-eb3ab44f6dcd.png)

Now edit a robot test, adding `Sleep  10 minutes` or so. Go to the ZMI in the new http://localhost:55001/ Plone Site created for robots. Go to the ZMI and portalquickinstaller. You'll see:

![selecao_125](https://cloud.githubusercontent.com/assets/8203476/13155896/24978756-d667-11e5-9ec9-e292d890e430.png)

If you go to a bunch of tools (or error_log), you always get redirected to

> There are no transactions that can be undone. 

and if you try to go into any created content, you get an error.

We have tested this in other packages (like collective.messagesviewlet) and the package doesn't have this Plone Site difference behavior when using `bin/test`. What from collective.cover is causing this difference? This problem is holding us back in a test.

hvelarde:
my guess is this line: https://github.com/collective/collective.cover/blob/master/src/collective/cover/configure.zcml#L15

we should replace that by an explicit list.

on the other side, collective.cover uses Grok and all the magic included with it; we should get rid of that also.

idgserpro:
`bin/instance` and bin/test uses the same configure.zcml, but only in `bin/test` we have this erratic behavior. Do you think there's some magic happening there that only applies to `bin/test`? This is really hard to debug. :( Do you know if there's anyone from Plone Framework team that could help us in this manner? Maybe other packages have the same problem we're facing right now.

hvelarde:
yes, they use the same `configure.zcml` but it behaves in a different way: the line I pointed out is relevant because tests [explicitly disable autoinclude feature set](https://pypi.python.org/pypi/plone.app.testing/4.2.5#layer-reference).

ask on #plone to see if anybody has experienced issues like that.

PR #587, opened 2016-02-19
Shows message if an exception is thrown (fix #581)

### LINKED ISSUE #581: If an exception is thrown in a tile, @@updatetilecontent gives an error 500 and nothing is shown to the user
This is difficult to reproduce, since we're using a portuguese tile component in a complex setup that has this behavior, so we'll try to simulate it editing some of collective.cover implementation files to easy the test:

1 - Clone this repository (we tested using commit 6d615de1b3a056a336501e1d61354114c68e8585, before 1.0a13 release);
2 - ```python bootstrap.py``` and ```bin/buildout```;
3 - Create a simple cover and add a simple tile;
4 - Stop Plone;
5 - Edit **src/collective/cover/tiles/basic.py** and add a bunch of

```python
raise Exception("Testing cover handling of tiles brokenness")
```

to all methods (or choose any method that's called in compose view when you select a content to add);

6 - Start Plone again;
7 - You should see:

selecao_113

This is not a helpful message, but at least the cover isn't broken (similar message when you have a broken portlet);
8 - Go to compose;
9 - Try to add a new item to the basic tile that throwns the exception;
10 - You get:

selecao_112

Infinite loading and a 500 Internal Server error, so it's not possible to the user to know if an error occured or if the server is slow. When you click compose again, you see that nothing was saved.

I'm not familiar with this component source code, neither it's javascript implementation, but I think a message should be shown to the user telling an error on the server happened: don't know the best place to show it. Maybe in the place of the loading icon?

idgserpro:
When composing, if an error occurs in a tile and an exception is thrown,
nothing is shown to the user. This commit fixes this behavior, closing

    https://github.com/collective/collective.cover/issues/581

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.0a13 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
+- Shows message to user if an exception is thrown in a tile (closes `#581`_).
+  [idgserpro]
+
 - Fix date format in collection tiles (closes `#584`_).
   [tcurvelo]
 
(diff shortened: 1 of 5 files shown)


idgserpro:
@rodfersou this fixes https://github.com/collective/collective.cover/issues/581. @hvelarde this is what motivated https://github.com/collective/collective.cover/issues/586

PR #600, opened 2016-03-08
Do not use the calendar tool to discover Event-like objects as it was removed on Plone 5

hvelarde:


hvelarde:
@idgserpro @winstonf88 please review as this could have implications for brasil.gov.agenda content type.

PR #601, opened 2016-03-08
Remove 'magic' around plone.app.stagingbehavior dependency

hvelarde:
We simply should not depend on it in any way; that's most likely the work of a policy package.


hvelarde:
@idgserpro @winstonf88 FYI

PR #603, opened 2016-03-09
Remove Grok Dependency for Cover View.

l34marr:
Another small step toward removing Grok dependency.

For the record: By first try, I assume all classes in browser/cover.py are equally Browser Views. Then I run into an error:
`LocationError: (<Products.Five.metaclass.SimpleViewClass from ...)`
that leads me to https://github.com/collective/collective.nitf/issues/80
So I try some DefaultView instead of BrowserView. It seems working, but I might need advice or insight.

### DIFF
--- a/src/collective/cover/browser/configure.zcml
+++ b/src/collective/cover/browser/configure.zcml
@@ -7,10 +7,113 @@
     >
 
   <browser:page
+      name="view"
+      for="..interfaces.ICover"
+      permission="zope2.View"
+      class=".cover.View"
+      template="templates/view.pt"
+      />
+
+  <browser:page
+      name="standard"
+      for="..interfaces.ICover"
+      permission="zope2.View"
+      class=".cover.Standard"
+      template="templates/standard.pt"
+      />
+
+  <browser:page
+      name="addctwidget"
+      for="..interfaces.ICover"
+      permission="cmf.ModifyPortalContent"
+      class=".cover.AddCTWidget"
+      />
+
+  <browser:page
+      name="addtilewidget"
+      for="..interfaces.ICover"
+      permission="cmf.ModifyPortalContent"
+      class=".cover.AddTileWidget"
+      />
+
+  <browser:page
+      name="setwidgetmap"
+      for="..interfaces.ICover"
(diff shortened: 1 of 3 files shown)


hvelarde:
@l34marr awesome! there is a small issue with code analysis; could you please fix it?

hvelarde:
@idgserpro @winstonf88 @rodfersou I need help to review with this one.

l34marr:
I know little about robotframework testing. Hints are appreciated so that I can learn some more.

hvelarde:
do you mind to make a rebase first?

l34marr:
I try rebase first and see http://stackoverflow.com/questions/19016698/git-branch-diverged-after-rebase so instead I click the "Update branch" button on GitHub. Hopefully it works for this case.

src/collective/cover/browser/configure.zcml
```diff
@@ -7,10 +7,113 @@
...
+  <browser:page
+      name="removeitemfromlisttile"
+      for="..interfaces.ICover"
+      permission="cmf.ModifyPortalContent"
+      class=".cover.RemoveItemFromListTile"
+      />
+
+  <browser:page
+      name="deletetile"
+      for="..interfaces.ICover"
```
idgserpro: @hvelarde do we have any best practices in zcml when naming for and class attributes? Should we put the whole path, or like this? This just hit my mind.

hvelarde: I have no idea, but it should be good to do so as packages as [z3c.dependencychecker](http://pypi.python.org/pypi/z3c.dependencychecker) can read ZCML files, IIRC.

hvelarde:
@l34marr most of the time is just `git rebase master` and solve any conflicts in the middle; you will need to use `git reset --hard origin/remove-grok-view` to update your clone now.

l34marr:
I try `git rebase` but things become even complicated, so I create another pull #634 to supersede this one. Hopefully this will make things easier to move on. If that works, you can close this one accordingly.

hvelarde:
ok, I'm closing this then.

PR #610, opened 2016-03-21
Disable plone.app.widget test matrix

### LINKED ISSUE #580: Fix RF tests with plone.app.widgets installed
Following tests are failing after installing plone.app.widgets:

* [x] Test Collection Tile (`ValueError: Element locator 'xpath=//select[@name="addindex"]' did not match any elements.`)
* [ ] Test Content Body Tile (`ValueError: Element locator 'id=text' did not match any elements.`)
* [ ] Test RichTextTile is Searchable (`ElementNotVisibleException: Message: Element is not currently visible and so may not be interacted with Stacktrace:`)

For more information, see:

* https://travis-ci.org/collective/collective.cover/jobs/103581538
* https://s3-sa-east-1.amazonaws.com/collective-cover/collective/collective.cover/2517/2517.3/parts/test/robot_log.html

rodfersou:
closes #580

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,6 @@ notifications:
       - "Build details : %{build_url}"
       - "Robot Test Report for Plone 4.2: http://s3-sa-east-1.amazonaws.com/collective-cover/collective/collective.cover/%{build_number}/%{build_number}.1/parts/test/robot_log.html"
       - "Robot Test Report for Plone 4.3: http://s3-sa-east-1.amazonaws.com/collective-cover/collective/collective.cover/%{build_number}/%{build_number}.2/parts/test/robot_log.html"
-      - "Robot Test Report for jQuery 1.8: http://s3-sa-east-1.amazonaws.com/collective-cover/collective/collective.cover/%{build_number}/%{build_number}.3/parts/test/robot_log.html"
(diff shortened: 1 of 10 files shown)


rodfersou:
A side effect of add ``plone.app.widgets`` is that it is deprecating ``plone.app.jquery`` resource, and using ``jQuery`` that is inside ``plone.app.widgets``.

But this version of ``jQuery`` is prepared to work just with ``AMD`` (requirejs), and don't expose anymore ``window.jQuery`` object, what makes our Plone site break as we have many plugins that require ``window.jQuery`` object.

We can change ``plone.app.jquery`` bundle from ``deprecated`` to ``default`` to work again as a fast workaround into ``portal_javascripts``.

A better fix would be to make version 1.8.0 version of ``plone.app.widgets`` use jquery from ``plone.app.jquery`` instead of replace the existing jQuery object, as mentionated in [this comment](https://github.com/plone/plone.app.widgets/commit/18fc9af6c440800cce7bbbb99c26d17f2564e45b#commitcomment-16782973).

This is not hard to fix, but we are not using ``plone.app.widgets`` in our use cases, so we are disabling ``plone.app.widgets`` into our test matrix now.

PR #612, opened 2016-03-22
Issue 611 - Fix rich text tile edit on Plone 4.2

### LINKED ISSUE #611: Can't close edit tile modal on Plone 4.2
To reproduce it:

* create a new cover object in Plone 4.2
* add an embed tile on Layout
* go to compose and edit the tile

You can save, but the modal remains open; cancel button doesn't work neither.

There is an `TypeError: document.getElementById(...) is null` error on tiny_mce_init.js:7:37.

This happens on banner, basic, collection, embed and file tiles as you can see in:

https://travis-ci.org/collective/collective.cover/jobs/117478495

Seems to be related with #543.

rodfersou:
closes #611

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ env:
     - PLONE_VERSION=5.0
 matrix:
   allow_failures:
-    - env: PLONE_VERSION=4.2
     - env: PLONE_VERSION=5.0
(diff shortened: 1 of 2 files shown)


rodfersou:
I think this fix should not have changelog entry.

PR #614, opened 2016-03-28
Enforce usage of plone.app.blocks 2.2.1 to avoid issues with tiles breaking the whole cover page

hvelarde:
CC @idgserpro


idgserpro:
@hvelarde do you can translate the [message](https://github.com/plone/plone.app.blocks/blob/2.x/plone/app/blocks/tiles.py#L17) to the portuguese?

hvelarde:
if a tile breaks, the rest of the cover page will render without problems; please help testing it.

a RF test would be appreciated.

PR #617, opened 2016-03-31
Quality Assurance

hvelarde:
Update code analysis and fix (almost) all errors.


hvelarde:
@idgserpro do you mind taking care of reviewing this?

idgserpro:
@hvelarde mock 1.3 requires [setuptools>=17.1](https://github.com/testing-cabal/mock/blob/1.3.0/setup.py#L13) but Plone 4.2 setuptools is [0.6c11](https://github.com/plone/buildout.coredev/blob/4.2/versions.cfg#L20). I think only install because used the cache.

hvelarde:
the rest in fine?

idgserpro:
Yes, the rest is ok.

:+1:

idgserpro:
Maybe should be open an issue to fix blind exceptions in the future.

PR #630, opened 2016-06-15
Pin plone.tiles to 1.4 

fredvd:
as it was unpinned and the buildout/tests is fetching 1.5.2/latest version from pypi.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,6 +122,7 @@ Edit your buildout.cfg and add add the following to it:
     collective.js.bootstrap = 2.3.1.1
     plone.app.blocks = 2.2.1
     plone.app.tiles = 1.0.2
+    plone.tiles = 1.5.2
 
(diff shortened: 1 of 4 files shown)


hvelarde:
I think you have to pin it also on the other versions files.

fredvd:
I first wanted to test with the 4.3 and made the pull request to get travis-ci.org working.

Runing bin/test on master locally (which hasn't been touched for a month) currently breaks a number of robot tests for me.

fredvd:
I've restarted the Plone 4.2 test build on travis because it failed, but it's strange: I've only updated versions-4.3.x.cfg, so the Plone 4.2 test build shouldn't be affected and pass like the last succesful build from a month ago on master: https://travis-ci.org/collective/collective.cover/builds/127809733

hvelarde:
I have to take a look later to this to see why is failing.

PR #636, opened 2016-07-21
Fix keyword to deal with plone4.csrffixes delays

hvelarde:


idgserpro:
@hvelarde done. You can merge?

src/collective/cover/tests/cover.robot
```diff
@@ -44,7 +44,8 @@ Setup Cover Test Case
 Click Add Cover
     Open Add New Menu
     Click Link  css=a#collective-cover-content
-    Page Should Contain  Add Cover
+    Wait Until Page Contains  Add Cover
+    Sleep  1s  Wait for js
```
hvelarde: @rodfersou can we do this running some JS code? I don't want to use `Sleep`

idgserpro: @hvelarde only works well when the plone 4.csrffixes is installed.

idgserpro: > can we do this running some JS code?

Isn't the cover but plone4.csrffixes.

rodfersou: @hvelarde Why don't we use this instead? https://github.com/collective/collective.cover/blob/master/src/collective/cover/tests/test_calendar_tile.robot#L46

idgserpro: The problem here is that the robot can't click in [Save](https://github.com/collective/collective.cover/blob/master/src/collective/cover/tests/cover.robot#L56) while the js of plone4.csrffixes not do your job. 

@hvelarde why not use Sleep? The tests already takes long. Increase 1 second will not make a difference. We have Sleep in the code:

https://github.com/collective/collective.cover/blob/985d0faafdd3b3401c2ba56d5c0c2d8aaac2b48e/src/collective/cover/tests/test_carousel_tile.robot#L118

hvelarde: `Sleep` should be avoided as it may also fail if the process takes more than 1 second; could you please check if there is another condition that we could check to make sure plone4.csrffixes did its job?

rodfersou:
@idgserpro how can we test this change I did?

src/collective/cover/tests/cover.robot
```diff
@@ -43,8 +43,8 @@ Setup Cover Test Case
 
 Click Add Cover
     Open Add New Menu
-    Click Link  css=a#collective-cover-content
-    Page Should Contain  Add Cover
+    Wait Until Keyword Succeeds  5 sec  1 sec  Click Link  css=a#collective-cover-content
```
idgserpro: This isn't a way to sleep well?

@hvelarde if it works that way you agree?

idgserpro:
What happens is that when the plone4.csrfixes is installed, [Input tex](https://github.com/collective/collective.cover/blob/66ce73a3c9b43ee1dd619e915d36bf8fac9f15ca/src/collective/cover/tests/cover.robot#L53) did not work and the title was not filled. The keyword ```Input text for sure``` fixed the problem.

@hvelarde you can merge?

hvelarde:
@idgserpro where is the documentation on that keyword?

idgserpro:
@hvelarde check https://github.com/plone/plone.app.robotframework/blob/4ed11100c45351e1488f593da2b61e1b31a19b55/src/plone/app/robotframework/selenium.robot#L96

idgserpro:
@hvelarde you removed ```Input text for sure```?

https://github.com/collective/collective.cover/pull/636/commits/bcdbfb63a2622733055453803e435fc3ccad4709

hvelarde:
@idgserpro sorry, made a mistake while rebasing from an outdated branch; do you mind to fix it again?

idgserpro:
@hvelarde done.

PR #650, opened 2016-08-18
Remove dependency on plone.directives.form

l34marr:
I guess plone.directives.form can be replaced by z3c.form, plone.supermodel and plone.autoform, so here is my take to remove its dependency. Hopefully this will get rid of grok dependency completely. A fresh manual installation shows it is working.


l34marr:
I know little about robot tests. Help needed for the remaining test errors.

src/collective/cover/behaviors/interfaces.py
```diff
@@ -25,10 +26,16 @@ class IRefresh(model.Schema):
...
 
-@form.validator(field=IRefresh['ttl'])
-def validate_ttl(value):
-    if value <= 0:
-        raise Invalid(_(u'Value must be greater than zero.'))
+class TimeToLiveValidator(validator.SimpleFieldValidator):
+
+    """z3c.form validator class for time to live"""
+
+    def validate(self, value):
```
hvelarde: @l34marr this validator is not being tested in any way; can you write a test?

hvelarde:
@idgserpro I need help with this also.

src/collective/cover/behaviors/interfaces.py
idgserpro: @hvelarde it seems to me correct. See: http://docs.plone.org/develop/addons/schema-driven-forms/customising-form-behaviour/validation.html#field-widget-validators

hvelarde:
@idgserpro awesome; I think we can make a new release now, do you agree?

idgserpro:
Yes. Remember bump version to 1.3

PR #653, opened 2016-08-18
Issue 633 - Remove dependency of portal_calendar tool

### LINKED ISSUE #633: Calendar tile does not work under Plone 5
The code of the calendar tile was copied from the calendar portlet and it depends on the `portal_calendar` tool; this tools was removed in Plone 5:

https://travis-ci.org/collective/collective.cover/jobs/141693521#L1392

We need to refactor this part of the code to allow compatibility.

Some ideas can be taken from #600.

rodfersou:
closes #633


src/collective/cover/tiles/calendar.py
```diff
@@ -169,12 +177,10 @@ def isToday(self, day):
...
 
     def getReviewStateString(self):
-        states = self.calendar.getCalendarStates()
-        return ''.join(map(lambda x: 'review_state={0}&amp;'.format(self.url_quote_plus(x)), states))
+        return 'review_state=published&amp;'
 
     def getEventTypes(self):
-        types = self.calendar.getCalendarTypes()
-        return ''.join(map(lambda x: 'Type={0}&amp;'.format(self.url_quote_plus(x)), types))
+        return 'Type=Event&amp;'
```
hvelarde: @rodfersou do not hard-code this; all content types providing `plone.event.interfaces.IEvent` must be treated as events.

see: https://community.plone.org/t/canonical-way-of-declaring-a-content-type-as-event-like/1735?u=hvelarde

src/collective/cover/tiles/calendar.py
```diff
@@ -169,12 +177,10 @@ def isToday(self, day):
             self.now[2] == day and self.now[1] == self.month and self.now[0] == self.year)
 
     def getReviewStateString(self):
-        states = self.calendar.getCalendarStates()
-        return ''.join(map(lambda x: 'review_state={0}&amp;'.format(self.url_quote_plus(x)), states))
+        return 'review_state=published&amp;'
```
hvelarde: @rodfersou do not hard-code this; all content types providing `plone.event.interfaces.IEvent` must be treated as events.

see: https://community.plone.org/t/canonical-way-of-declaring-a-content-type-as-event-like/1735?u=hvelarde

src/collective/cover/tiles/calendar.py
```diff
@@ -107,15 +124,6 @@ def getYearAndMonthToDisplay(self):
         year = request.get('year', None)
         month = request.get('month', None)
 
-        # Next get the data from the SESSION
-        if self.calendar.getUseSession():
```
hvelarde: @rodfersou I think you can not get rid of this.

src/collective/cover/tests/test_calendar_tile.py
```diff
@@ -10,10 +12,25 @@ class CalendarTileTestCase(TestTileMixin, unittest.TestCase):
...
         self.tile.id = u'test'
 
+        self.events = []
+        with api.env.adopt_roles(['Manager']):
+            for i in [3, 12, 21, 30]:
+                start = datetime(2016, 8, i, 10, 30)
+                end = datetime(2016, 8, i, 11, 30)
+                obj = api.content.create(
+                    container=self.portal, type='Event', id='e{0}'.format(i),
+                    startDate=start, endDate=end)
```
hvelarde: @rodfersou this is not going to work as Dexterity-based content types field names are different than Archetypes-based.

see: https://github.com/plone/plone.app.event/issues/209

src/collective/cover/tiles/calendar.py
```diff
@@ -1,9 +1,12 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
```
hvelarde: @rodfersou add a comment on why we are importing this.

src/collective/cover/tiles/calendar.py
rodfersou: why not? this calendar tile don't need to use session to keep the year and month as parameter. I'll just use GET parameters to do it.

src/collective/cover/tiles/calendar.py
rodfersou: @hvelarde I didn't finish the changes.  My first step is to remove `portal_calendar`, after I'll refactor the code and these methods will be removed.

src/collective/cover/tiles/calendar.py
rodfersou: @hvelarde I didn't finish the changes.  My first step is to remove `portal_calendar`, after I'll refactor the code and these methods will be removed.

src/collective/cover/tiles/calendar.py
```diff
@@ -56,31 +61,57 @@ def _setup(self):
...
-                    localized_date = [self._ts.ulocalized_time(cur_date, context=context, request=self.request)]
-                    day['eventstring'] = '\n'.join(localized_date + [
-                        ' {0}'.format(self.getEventString(e)) for e in day['eventslist']])
-                    day['date_string'] = '{0}-{1}-{2}'.format(year, month, daynumber)
+    def addEvents(self, day):
+        catalog = api.portal.get_tool('portal_catalog')
+        start = DateTime(self.year, self.month, day['day'])
+        end = start + 1
+        query = dict(
+            portal_type='Event', review_state='published',
```
hvelarde: @rodfersou as mentioned before, this can't be hard coded: the query must search for all contenty types providing the `IEvent` interface.

src/collective/cover/tiles/calendar.py
```diff
@@ -168,14 +184,6 @@ def isToday(self, day):
         return (
             self.now[2] == day and self.now[1] == self.month and self.now[0] == self.year)
 
-    def getReviewStateString(self):
```
hvelarde: @rodfersou these methods should not be removed; but fixed.

src/collective/cover/tiles/calendar.py
```diff
@@ -56,31 +61,57 @@ def _setup(self):
...
-                if daynumber == 0:
-                    continue
-                day['is_today'] = self.isToday(daynumber)
-                if day['event']:
-                    cur_date = DateTime(year, month, daynumber)
-                    localized_date = [self._ts.ulocalized_time(cur_date, context=context, request=self.request)]
-                    day['eventstring'] = '\n'.join(localized_date + [
-                        ' {0}'.format(self.getEventString(e)) for e in day['eventslist']])
-                    day['date_string'] = '{0}-{1}-{2}'.format(year, month, daynumber)
+    def addEvents(self, day):
```
hvelarde: @rodfersou please add a doc string.

src/collective/cover/tiles/calendar.py
```diff
@@ -56,31 +61,57 @@ def _setup(self):
...
+                start=brain.start.Time(), end=brain.end.Time(), title=brain.Title or brain.id)
+            day['eventslist'].append(event)
+            if 'eventstring' not in day:
+                localized_date = self._ts.ulocalized_time(
+                    start, context=self.context, request=self.request)
+                day['eventstring'] = localized_date
+            day['eventstring'] += '\n {0}'.format(self.getEventString(event))
+        return day
 
+    def getEventsForCalendar(self):
```
hvelarde: @rodfersou please add a doc string.

src/collective/cover/tiles/calendar.py
```diff
@@ -100,22 +131,12 @@ def getEventString(self, event):
         return eventstring
 
     def getYearAndMonthToDisplay(self):
```
hvelarde: @rodfersou please add a doc string.

src/collective/cover/tiles/calendar.py
```diff
@@ -56,31 +61,57 @@ def _setup(self):
...
-                    continue
-                day['is_today'] = self.isToday(daynumber)
-                if day['event']:
-                    cur_date = DateTime(year, month, daynumber)
-                    localized_date = [self._ts.ulocalized_time(cur_date, context=context, request=self.request)]
-                    day['eventstring'] = '\n'.join(localized_date + [
-                        ' {0}'.format(self.getEventString(e)) for e in day['eventslist']])
-                    day['date_string'] = '{0}-{1}-{2}'.format(year, month, daynumber)
+    def addEvents(self, day):
+        catalog = api.portal.get_tool('portal_catalog')
```
hvelarde: @rodfersou the catalog call must be outside of the loop or you are going to call it up to 5 times instead of just one.

src/collective/cover/tiles/calendar.py
```diff
@@ -169,11 +187,13 @@ def isToday(self, day):
             self.now[2] == day and self.now[1] == self.month and self.now[0] == self.year)
 
     def getReviewStateString(self):
-        states = self.calendar.getCalendarStates()
+        states = ['published']
         return ''.join(map(lambda x: 'review_state={0}&amp;'.format(self.url_quote_plus(x)), states))
 
     def getEventTypes(self):
```
hvelarde: @rodfersou this must be memoized.

src/collective/cover/tiles/calendar.py
```diff
@@ -169,11 +187,13 @@ def isToday(self, day):
...
-        types = self.calendar.getCalendarTypes()
+        catalog = api.portal.get_tool('portal_catalog')
```
hvelarde: @rodfersou this must be memoized.

src/collective/cover/tests/test_upgrades.py
```diff
@@ -524,8 +524,6 @@ def test_registrations(self):
         self.assertGreaterEqual(int(version), int(self.to_version))
         self.assertEqual(self._how_many_upgrades_to_do(), 4)
 
-    # FIXME: https://github.com/collective/collective.cover/issues/633
```
hvelarde: @rodfersou do not remove this or test will fail in Plone 5; only remove the FIXME comment line.

src/collective/cover/static/js/main.js
```diff
@@ -1,8 +1,51 @@
...
+    ajaxOptions.error = function(){
+      options.error();
+    }
+    ajaxOptions.data = options.data;
+    $.ajax(ajaxOptions);
+  }
+  $('body').undelegate('#calendar-next, #calendar-previous', 'click')
+           .delegate(
+             '.portletWrapper #calendar-next, ' +
+             '.portletWrapper #calendar-previous',
```
rodfersou: In this override, I just set the selector to target just portlets with the ``.portletWrapper`` class (maybe we should create a PR for ``plone.app.portlets`` package)

src/collective/cover/static/css/cover.css
```diff
@@ -249,30 +249,23 @@ table.invisible
 }
 
 /* Tile Calendar */
-.cover-calendar-tile .calendar-tile-header
+.cover-calendar-tile .portletCalendar
```
hvelarde: @rodfersou I think this is a bad idea.

@agnogueira comments?

src/collective/cover/static/js/main.js
hvelarde: @rodfersou maybe you can avoid that using a different seelctor.

src/collective/cover/static/css/cover.css
rodfersou: @hvelarde I have two options here.. or I do this way, or I have to customize 2 templates of both portlet.. you choose.

src/collective/cover/static/js/main.js
rodfersou: @hvelarde I have two options here.. or I do this way, or I have to customize 2 templates of both portlet.. you choose.

src/collective/cover/static/css/cover.css
rodfersou: @hvelarde I don't understand your concern here with this css code.. this selector is restrict to our calendar tile because of the ``.cover-calendar-tile`` class.  It will not  affect the portlet calendar.

src/collective/cover/static/js/main.js
rodfersou: @hvelarde the problem here is that I am getting the html from the portlet without change, this has the benefit to don't have to maintain this code, but the javascript from this portlet has this problem, the selector to refresh portlet is looking out of the portlet too.  It would be no problem if there were a ``<div>`` with class ``.portletWrapper`` and the portlet id as data attribute to make it work with the original javascript of the portlet, but we don't have it.

PR #657, opened 2016-09-10
Update Traditional Chinese Translation

l34marr:


hvelarde:
thanks! I'll make a release later today.

PR #658, opened 2016-09-12
Plone tiles 170

datakurre:
This is not a pull request. I just need to get Travis run.


datakurre:
Hmm... Travis does not seem to build right now. (Not commits, not pulls.) I return to this later.

PR #661, opened 2016-09-16
Fix backwards compatibility for BaseGrid

fredvd:
If you want to import BaseGrid, get it from the right location.


hvelarde:
thanks!

PR #663, opened 2016-09-16
Enforce usage of plone.app.tiles >= 1.1.0 to avoid the creation of zillions of empty blob files when using versioning

### LINKED ISSUE #532: Versioning creates zillions of empty blob files
When using [plone.app.iterate](https://pypi.python.org/pypi/plone.app.iterate), each version of a content object with persistent tiles with blobs, will create a new empty blob into file system for each found blob and I will also break the CMFEditions Portal ZVC based Histories Storage.

If you're using this feature in your site you have to take special attention to the number of free [inodes](https://en.wikipedia.org/wiki/Inode) in your file system, as you can run out of them; use the ``df -i`` command to check it:

```bash
$ df -i
Filesystem            Inodes   IUsed   IFree IUse% Mounted on
rootfs                    10485760 8929045 1556715   86% /
udev                       2057472     272 2057200    1% /dev
tmpfs                      2058962     204 2058758    1% /run
/dev/disk/by-label/DOROOT 10485760 8929045 1556715   86% /
tmpfs                      2058962       1 2058961    1% /run/lock
tmpfs                      2058962       5 2058957    1% /run/shm

```

hvelarde:
fixes #532

refs. https://github.com/plone/plone.app.tiles/pull/17


hvelarde:
@idgserpro @fredvd @mauritsvanrees @djowett @fulv FYI, this is huge!

fulv:
How backward compatible is plone.app.tiles >= 1.1.0?  I.e., can I use it
with older versions of c.cover?

hvelarde:
@fulv we've only seen issues with tests on the ESI feature, but I think nobody is using that.

PR #669, opened 2016-09-21
Use warnings built-in module instead of zope.deprecation

hvelarde:


hvelarde:
@mauritsvanrees thank you, again; do you have an example on how to achieve the warning on the old `BaseGrid` class?

mauritsvanrees:
No example. But one of the options in the deprecation guide probably does what you want: http://docs.plone.org/develop/styleguide/deprecation.html

idgserpro:
Just out of curiosity, whats the practical advantage of removing zope.deprecation? Just to remove a dependency? We're already in Plone, zope.deprecation is there, anyways. Or do we have plans of removing it from Plone? @mauritsvanrees?

hvelarde:
@idgserpro in fact there's no advantage; I was confused because the deprecation warning messages were not showing up locally; but the issue was a missing `deprecation-warnings = on`.

idgserpro:
I develop.cfg in UnifiedInstaller is [set to on](https://github.com/plone/Installers-UnifiedInstaller/blob/7a73e8b9d529f363a3b18d153a5b602eac989302/base_skeleton/develop.cfg).

PR #681, opened 2016-11-01
Don't show image on tile collection if image attribute is None

rodfersou:
related to: https://github.com/collective/collective.nitf/pull/179/files
closes https://github.com/collective/collective.nitf/issues/178

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.3b2 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- Don't show image on tile collection if image attribute is None.
+  [rodfersou]
+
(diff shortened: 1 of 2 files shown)


src/collective/cover/tests/test_collection_tile.py
```diff
@@ -76,6 +77,16 @@ def test_populate_tile_with_object_string(self):
         self.assertTrue(self.tile.data.get('footer'))
         self.assertEqual(self.tile.data.get('uuid'), IUUID(obj))
 
+    def test_populate_tile_without_image(self):
```
hvelarde: @rodfersou this test does not work: you have to create a test in collective.nitf for this, make the test fail and test with this branch to see if that fixes the issue.

rendering the tile must rise an exception and this is not happening here,

hvelarde:
I'm fixing the cause in collective.nitf.

PR #682, opened 2016-11-11
Fixed adding a 'more' link in list tiles.

mauritsvanrees:
Previously you could select an item to use as 'more' link, but it did not stick.
I checked previous versions: this worked in 1.0a12, but failed since 1.1b1.

I have expanded the robot test for the list tile.
For clarity let me give two screen shots made during this test.

First a screen shot of the filled-in list tile edit form:

screen shot 2016-11-11 at 18 42 53

And then the resulting view, annotated for clarity:

screen shot 2016-11-11 at 18 43 14 annotated

PR #683, opened 2016-11-16
Fixed too long line.

mauritsvanrees:
Goal: test Travis with a minor change.

Travis gives problems on pull request #682, where the tests pass, but the job hangs and gets killed. So let's check with a very minor change, to make sure that it really has nothing to do with the robot test that I added in that pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,11 @@ after_success:
 - bin/createcoverage -t "--layer=!Robot"
 - pip install coveralls
 - coveralls
+after_script:
+# Firefox complains: GConf-WARNING **: Got Disconnected from DBus.
+# And then it keeps hanging, causing a failed job over 30-50 minutes.
+# So we kill Firefox.
+- pkill firefox
 notifications:
   irc:
     on_success: change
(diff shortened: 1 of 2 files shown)


hvelarde:
seems it's working… awesome! do we really need to kill the Firefox process besides using the ESR version?

hvelarde:
yes, we need it: https://travis-ci.org/collective/collective.cover/jobs/176685533

hvelarde:
awesome! I'm adding those changes in the other PR to merge it.

mauritsvanrees:
There was one Travis failure even with killing, but that seemed a test that may have been unstable. I restarted that single job and it worked.

It could be that the latest ESR is not needed, and that killing is enough. But it seems good to test with a more recent Firefox.

About an hour ago I restarted a job from the other PR to see if things magically started working there, which would have meant it was all just temporary wrong behavior from Firefox on Travis. But that job ran the tests fine and then kept hanging as well. So it seems killing the process is now really needed.

Note that with normal `pkill` it still failed in one case, but changing it to`pkill -9` helped.

PR #690, opened 2016-12-11
Fix @@updatetilecontent view to avoid rendering outdated data

hvelarde:


src/collective/cover/browser/compose.py
```diff
@@ -50,10 +50,16 @@ def render(self):
...
+            # reinstantiate tile to update rendering with new data
+            tile = self.context.restrictedTraverse(path)
             return tile()
 
     def __call__(self):
         self.setup()
+        # avoid caching the response on intermediate proxies
+        # as this is a GET request, a misconfiguration could
+        # result on outdated content being rendered
+        self.request.response.setHeader('Cache-Control', 'no-cache')
```
hvelarde: @fredvd I was checking your previous commits about this and I was curious about the inclusion of `must-revalidate` in the `Cache-Control` header; AFAIK, that is only needed if the content is cached.

comments?

fredvd: @hvelarde I am not really an expert in all the different combinations of caching headers. I was extremely happy at the time to have found the cause of the outdated data at all but wasn't sure at first who was caching (the browser or an intermediate proxy, so I probably looked up some info back then and selected the headers I though most likely to fix the problem.

When debugging this I saw 'not modified' responses in IE10's developer tools network tab, which led me to believe Varnish was caching. Later I found out it was IE that was buggy: it was not sending Last Modified requests at all when no Cache-Control header was present (but still pretending LM checking in the network tab). 

So maybe must-revalidate is not necessary anymore if you use no-cache and correct timeouts, but I can't really get a to a conclusion from these posts I just found:

http://stackoverflow.com/questions/1046966/whats-the-difference-between-cache-control-max-age-0-and-no-cache

http://stackoverflow.com/questions/18148884/difference-between-no-cache-and-must-revalidate

hvelarde: thanks!

PR #694, opened 2016-12-22
Issue 637 - Convert data back to string when drop object with image

### LINKED ISSUE #637: Can't edit basic tile with Dexterity-based content types
Plone 4.3.10 with plone.app.contenttypes 1.1.1 installed:
- upload an Image or create a News Item with an image
- drop the item in a basic tile
- edit the tile
- save it

the following validation error is shown (`Wrong contained type`):

selection_073

this was not happening before with Plone 4.3.9.

rodfersou:
closes #637


rodfersou:
How should be the test of this fix?

hvelarde:
just add a test that replaces the image on the basic tile with another image; I think the test can be inside the current basic tile test so we don't need to add another file.

PR #698, opened 2017-04-04
Update CI environment

hvelarde:


versions-4.2.x.cfg
```diff
@@ -32,3 +31,6 @@ plone.tiles = 1.8.0
 # for testing
 funcsigs = 0.4
 mock = 1.0.1
+
+# XXX: https://github.com/plone/plone.api/issues/364
+plone.api = 1.6
```
idgserpro: Is it not necessary to put in alphabetical order?

versions-4.3.x.cfg
```diff
@@ -12,3 +21,6 @@ plone.tiles = 1.8.0
 
 # for testing compatibility with newer versions of jQuery
 plone.app.jquery = 1.7.2
+
+# XXX: https://github.com/plone/plone.api/issues/364
+plone.api = 1.6
```
idgserpro: Is it not necessary to put in alphabetical order?

versions-4.3.x.cfg
```diff
@@ -1,4 +1,13 @@
 [buildout]
+# FIXME: there are pending issues on all versions after 4.3.11
```
idgserpro: I think it would be interesting to put a reference to: https://github.com/plone/plone.app.testing/issues/39

versions-4.2.x.cfg
hvelarde: I don't think so; this will be removed more sooner than later.

versions-4.3.x.cfg
hvelarde: I don't think so; this will be removed more sooner than later.

versions-4.3.x.cfg
hvelarde: ok

idgserpro:
Ok now :+1:

hvelarde:
I'll add a couple of PR to test with latest Plone 4.3 and plone.api.

hvelarde:
done! #699 and #700.

PR #699, opened 2017-04-04
Use always latest version of plone.api

hvelarde:
refs. https://github.com/plone/plone.api/issues/364
refs. https://github.com/plone/plone.api/issues/413


hvelarde:
@idgserpro I need help here also; be aware of https://github.com/plone/plone.app.contenttypes/issues/465.

idgserpro:
Does your workaround not work because of plone/plone.app.contenttypes#465?

hvelarde:
@idgserpro I really don't understand what's going on here: my guess is that the overly complicated architecture of plone.app.contenttypes making all content types depend on behavior is probably breaking things here.

it could be possible trying to access `event.start` attribute is not working without using an adapter, which really, really sucks.

versions-4.3.x.cfg
```diff
@@ -14,5 +14,5 @@ plone.tiles = 2.1
 # XXX: PFG tile is deprecated and will be removed in collective.cover 3
 Products.PloneFormGen = <1.8.0.alpha1
 
-# XXX: https://github.com/plone/plone.api/issues/364
-plone.api = 1.6
+# XXX: must be clean up after new Plone versions
+plone.api = >=1.8.6
```
idgserpro: Why not pin the specific version?

hvelarde: we'll remove this later, for now we can test on latest version; anyway seems the release is a brown bag: buildout is failing with it:

```console
$ bin/buildout 
Develop: '/home/hvelarde/collective/cover/.'
/home/hvelarde/.buildout/eggs/setuptools-26.1.1-py2.7.egg/pkg_resources/__init__.py:188: RuntimeWarning: You have iterated over the result of pkg_resources.parse_version. This is a legacy behavior which is inconsistent with the new version class introduced in setuptools 8.0. In most cases, conversion to a tuple is unnecessary. For comparison of versions, sort the Version instances directly. If you have another use case requiring the tuple, please file a bug with the setuptools project describing that need.
  stacklevel=1,
Unused options for buildout: 'show-picked-versions' 'package-min-coverage'.
Installing instance.
Getting distribution for 'plone.api==1.8.6'.
While:
  Installing instance.
  Getting distribution for 'plone.api==1.8.6'.
Error: Couldn't find a distribution for 'plone.api==1.8.6'.
```

CC @gforcada

hvelarde: now I see: it's 1.8.5 indeed.

PR #700, opened 2017-04-04
Use latest version of Plone 4.3

### LINKED ISSUE #759: Robot Framework tests fail with Plone > 4.3.11
After upgrading to Plone 4.3.12 all Robot Framework tests started failing as hown in the following screenshot:

selenium-screenshot-1

as it can be seen, there's a new required field being shown in the form that normally doesn't appear; this field came from a [XML schema](https://github.com/collective/collective.cover/blob/1.6b2/src/collective/cover/models/cover.xml#L19-L26) like this:

```xml
<field name="cover_layout" type="zope.schema.Text"
       form:omitted="z3c.form.interfaces.IForm:true
                     z3c.form.interfaces.IAddForm:true
                     z3c.form.interfaces.IEditForm:true">
    <description i18n:translate=""
      >The layout to be used to render groups and tiles</description>
    <title i18n:translate="">Cover Layout</title>
</field>
```

as it can be seen, this field should be omitted on `z3c.form.interfaces.IAddForm` but that's not happening and I don't know why.

it's important to note that this happens only while running tests; the field is normally hidden when accessing the add form in a browser.

IMO this field must be redefined as `readonly` instead of being hidden all over the place.

hvelarde:
closes #759 as a workaround

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,8 @@ There's a frood who really knows where his towel is.
 1.6b3 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
-- Nothing changed yet.
+- Explicitly declare ``cover_layout`` field in content type schema as ``readonly``;
+  Robot Framework tests pass again with latest version of Plone 4.3 (fixes `#759 <https://github.com/collective/collective.cover/issues/759>`_).
 
 
 1.6b2 (2017-09-01)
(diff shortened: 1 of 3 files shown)


hvelarde:
@idgserpro I need your review on this, all tests are passing again.

idgserpro:
LGTM. But I still think https://github.com/plone/plone.app.testing/issues/39 must be solved to avoid problems in the future.

This may mask some error in tests.

PR #704, opened 2017-04-05
Use latest versions of all Deco related packages

hvelarde:
we need to keep Deco related packages up to date to avoid surprises.

### DIFF
--- a/versions-4.3.x.cfg
+++ b/versions-4.3.x.cfg
@@ -15,10 +15,9 @@ test-eggs =
 
 [versions]
 collective.js.bootstrap = 2.3.1.1
-plone.app.blocks = 3.1.0
-plone.app.drafts = 1.0
-plone.app.tiles = 1.1.0
-plone.tiles = 1.8.0
+plone.app.blocks = 4.0.6
+plone.app.tiles = 3.0.3
+plone.tiles = 2.0.0b2
 
 # for testing compatibility with newer versions of jQuery
 plone.app.jquery = 1.7.2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,14 @@ There's a frood who really knows where his towel is.
 1.4b2 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+.. Warning::
+    If you are upgrading plone.app.tiles note that latests versions of this package no longer depend on plone.app.drafts.
+    You should explicitly add plone.app.drafts to the `eggs` part of your buildout configuration to avoid issues.
+    You can safely unistall plone.app.drafts after that, if you are not using it.
+
+- Update recommended versions of Blocks dependencies to keep in sync with curren Mosaic development.
+  [hvelarde]
+
 - Fix order of uuids of sorted function in ListTile's 'results' method.
   [idgserpro]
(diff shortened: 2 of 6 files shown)


datakurre:
@hvelarde Regarding collective.cover there should be no breaking changes, including the changes, which I'm currently planning for plone.tiles 2.0 final and next plone.app.blocks.

Btw, what requires you to depend on plone.app.drafts? Or does it work well enough with cover? (Also drafting needs work. Currently drafting stores unsaved changes as a side effect of inline-validation. That will probably be replaced with more explicit API in the future, which related pattern for normal edit forms and integration in Mosaic editor.)

hvelarde:
@datakurre I was thinking plone.app.drafts was a dependency of plone.app.tiles and now I see it's only a test dependency; I'll remove it:

https://github.com/plone/plone.app.tiles/blob/master/setup.py#L67

at some point I want to get rid of the layout editor in here to use yours.

hvelarde:
@datakurre mmm, seems we have a problem: plone.app.tiles depended on plone.app.drafts so I can't just update because I'm getting:

```python
AttributeError: type object 'IDraftStorage' has no attribute '__iro__' 
```

we need to add plone.app.drafts to the buildout and uninstall it; we'll need to add a note to the changelog.

idgserpro:
@datakurre is not this a problem that plone.app.tiles would have deal with? With a BBB or something?

versions-4.3.x.cfg
```diff
@@ -15,10 +15,9 @@ test-eggs =
...
-plone.app.blocks = 3.1.0
-plone.app.drafts = 1.0
-plone.app.tiles = 1.1.0
-plone.tiles = 1.8.0
+plone.app.blocks = 4.0.6
+plone.app.tiles = 3.0.3
+plone.tiles = 2.0.0b2
 
 # for testing compatibility with newer versions of jQuery
 plone.app.jquery = 1.7.2
```
idgserpro: Plone already [pin plone.app.jquery](https://github.com/plone/buildout.coredev/blob/4.3/versions.cfg#L166). I think it can be removed and treated [this](https://github.com/collective/collective.cover/blob/master/.travis.yml#L29) of another way.

datakurre:
@hvelarde Ok, so drafts requires uninstall profile then. There's persistent utility (portal_draftstorage), which gets installed with drafts.

hvelarde:
I need to check then if that persistent utility is gone; I'll come back with the results.

versions-4.3.x.cfg
hvelarde: this is being used for tests as you can read in the line above; I'll refactor it soon.

hvelarde:
@datakurre I uninstalled plone.app.drafts, removed the egg from the buildout configurtion; reran buildout and restarted the instance: no issues, so seems safe to remove it.

@idgserpro do you think is fine to just add a note to the changelog stating that you must add plone.app.drafts to the eggs and manually uninstall it if not being used it, or do you want me to do anything else?

idgserpro:
@hvelarde since ```plone.app.draft``` isn't a direct dependency of cover, I think only the documentation is fine.

PR #711, opened 2017-05-09
Issue 710 - Review contentchooser events to happen just at compose tab

### LINKED ISSUE #710: In Plone 4 all tiles can be dragged in the view template (have ui-draggable/droppable)
After installing 1.4b1 in a custom project and verifying the same behaviour in the collective.cover buildout: all outer tile divs now have ui-draggable and ui-droppable classes. 

When you view a cover page you can now no longer select any text inside a cover as you activate the draggable behaviour and visitors can drag around tiles around the page (that will snap back as soon they release the mouse button).

A quick search through the collective.cover code doesn't show any ui-draggable/droppable in templates or javascript, so I suspect this regression is coming from a higher pinned support package, plone.app.tiles or plone.app.blocks .

rodfersou:
closes #710


hvelarde:
we're going to work on that probably next week.

PR #714, opened 2017-05-20
Enable collection and folders to be listed at tile list; Tile carousel don't accept folder anymore.

### LINKED ISSUE #713: list tiles can no longer link to collections
Plone 4.3.14,  plone.app.contenttypes 1.1.2

With the drag and drop feature between tiles (https://github.com/collective/collective.cover/issues/487) another feature not described in the issue was added: if you drop a collection on a list tile, all the items in the collection will be added as individual items. I found out about this from the end-user.rst documentation where the feature is described:

>If you drop a folder or a collection into this tile, it will take every item into the folder or into the collection and insert it until it reaches the max items specified into configuration.

However it does not work as described when I drop a folder on a list tile, at least not with plone.app.contenttypes (default dexterity CT's) installed. Then the folder itself is added. **(1)**

Very fancy, but how do I link to a collection itself in a a list tile? This is no longer possible but was possible before, adding to the confusion that I do see links to collections in existing list tiles in our sites, but these were probably added before this was released. 

**(1)** The only work around I've found is to put the collection you want to link to in a folder, set the collection as the default view and drop the folder on the list tile in the compose screen. Lucky me that the end-user description for dropping folders on list tiles is not correct.

rodfersou:
closes #713


src/collective/cover/tiles/list.py
```diff
@@ -379,17 +406,6 @@ def get_title_tag(self, item):
 
 
 @implementer(ICoverUIDsProvider)
-class CollectionUIDsProvider(object):
```
rodfersou: should deprecate this instead of remove then just change the UIDs returned?

rodfersou:
Still lot's of broken tests to look at...

src/collective/cover/tiles/list.py
hvelarde: just remove all of them and document that on the changelog.

src/collective/cover/tiles/list.py
```diff
@@ -162,15 +162,14 @@ def set_limit(self):
                 self.limit = int(field.get('size', self.limit))
 
     def populate_with_object(self, obj):
-        """ Add an object to the list of items
+        """ Add an object to the list of items.
+        If object is a collection, replace items.
```
rodfersou: need to fix this docstring (back to original)

rodfersou:
After merge these pull requests we can get rid of the adapters and should pin PFG version here:

 * https://github.com/smcmahon/Products.PloneFormGen/pull/195
 * https://github.com/smcmahon/Products.PloneFormGen/pull/196

idgserpro:
This is a change in behavior. I think the version should be bumped to 1.5.

Users are used to adding a container and the 'magic' of populating with container's children. IMHO, a warning in the CHANGES.rst would be nice as well.

fredvd:
@idgserpro I feel your pain. In this case we just experienced the opposite change in behavior: users were used to adding a collection or folder to a list tile as reference and suddenly weren't able to do so anymore.  It's a difficult catch here. Maybe the mockup that @rodfersou made for modifying the action can be helpful here so we can support both use cases. 

But do your users also expect and depend on the unpacking of items for list tiles?

hvelarde:
@idgserpro I agree, the situation is that currently there's no way to add a collection or a folder to a list tile (which is a bug); after this fix, users used to to what you describe will still be able to populate the tile on an item base (which only implies a little more work).

idgserpro:
That's why we're discussing in https://github.com/collective/collective.cover/issues/716: in one hand, you have users wanting the behavior that @fredvd is asking, and in our case we already have the behavior of adding the children uuids of a container if it's a list/inherit from a list.

Having https://github.com/collective/collective.cover/issues/716 done, tiles that inherit from ListTile can override `populate_with_object` adding this behavior. What we're asking is not for you to don't merge this PR, but ease the path of people that still use this behavior.

src/collective/cover/tiles/carousel.py
```diff
@@ -66,12 +66,21 @@ def populate_with_object(self, obj):
...
+            uuids = [i.UID for i in obj.queryCatalog()]
+        else:
+            uuids = [self.get_uuid(obj)]
         # accept just elements with a lead image
         uuids = [i for i in uuids if self._has_image_field(uuidToObject(i))]
         if uuids:
             self.populate_with_uuids(uuids)
 
+    @view.memoize
+    def accepted_ct(self):
```
hvelarde: this tile should in fact accept any content type as even Dexterity folders can have an image field using a behavior; the validation must be done in the `populate_with_object()` method.

src/collective/cover/tiles/carousel.py
```diff
@@ -66,12 +66,21 @@ def populate_with_object(self, obj):
         :type uuids: List of strings
         """
         super(ListTile, self).populate_with_object(obj)  # check permission
-        uuids = ICoverUIDsProvider(obj).getUIDs()
+        if obj.portal_type in ('Topic', 'Collection'):
```
hvelarde: Topic has been deprecated since eons.

src/collective/cover/tiles/carousel.py
idgserpro: > the validation must be done in the populate_with_object() method.

How should a message be shown to the user if there's an invalid `portal_type` inside a container being validated by `populate_with_object`? Maybe this approach can be used in brasil.gov.tiles as well.

src/collective/cover/tests/test_carousel_tile.py
```diff
@@ -37,7 +39,14 @@ def test_default_configuration(self):
         self.assertTrue(self.tile.is_editable)
 
     def test_accepted_content_types(self):
-        self.assertEqual(self.tile.accepted_ct(), ALL_CONTENT_TYPES)
+        # Add folder to default accepted_ct
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ICoverSettings)
+        settings.searchable_content_types += ['Folder']
```
hvelarde: `searchable_content_types` has nothing to do with accepted content types.

src/collective/cover/tiles/carousel.py
idgserpro: ...or should "invalid" or "not expected" types be silently "swallowed" in `populate_with_object`?

src/collective/cover/tests/test_carousel_tile.py
idgserpro: Actually, if you inherit from base, you don't need to override `accepted_ct` to add a new valid accepted type, you can do like @rodfersou did:

https://github.com/collective/collective.cover/blob/2956dfc5a5c3c87af48efd679e0ce3391dca7ed9/src/collective/cover/tiles/base.py#L160

Is a matter of "taste", I suppose.

hvelarde:
@fredvd @smcmahon I made some changes according your comments; would you mind to make a code review? please take some special attention to the part with the deprecation warnings.

hvelarde:
I fixed the deprecation warnings and now everything seems to be fine.

src/collective/cover/tiles/carousel.py
```diff
@@ -66,7 +65,10 @@ def populate_with_object(self, obj):
         :type uuids: List of strings
         """
         super(ListTile, self).populate_with_object(obj)  # check permission
-        uuids = ICoverUIDsProvider(obj).getUIDs()
+        if obj.portal_type == 'Collection':
```
idgserpro: Shouldn't this be `if obj.portal_type in ['Collection', 'Folder']:` instead?

src/collective/cover/tiles/carousel.py
idgserpro: I guess you already answer in https://github.com/collective/collective.cover/blob/e23fed6b55222c54e5f16cb8631a74faddf56616/src/collective/cover/tiles/carousel.py#L71

src/collective/cover/tiles/carousel.py
hvelarde: no, folders are not going to populate the carousel; do you have a use case for that?

hvelarde:
@idgserpro deprecation warning are only for https://github.com/smcmahon/Products.PloneFormGen/issues/194 and we're keeping the code just until that is fixed; we're not aware of other people using it.

I'll bump the version at release time.

buildout.cfg
```diff
@@ -29,7 +29,7 @@ clean-lines-exclude = ${buildout:directory}/src/collective/cover/static/js/vendo
 csslint = True
 csslint-bin = bin/csslint
 debug-statements = True
-flake8-ignore = B901,D001,E501,P001,T000
+flake8-ignore = B901,D001,E501,P001,T000,Q000
```
idgserpro: Ignore Q000 it's no longer necessary: https://github.com/zheller/flake8-quotes/issues/55

src/collective/cover/tiles/carousel.py
idgserpro: All IDGB tiles that accept a Collection to do this "magic", accept a Folder as well. We know brasil.gov.tiles and collective.cover are completely different projects, but we thought that since you don't limit "Folder" in `accepted_ct` in carousel, it would make sense to have this feature for both Collection and Folder.

hvelarde: I don't limit it because is ignored anyway as it has no image field, but you can activate the lead image behavior on a Dexterity-based folder and then the folder will be shown in the gallery. I'm against most "magic" because it only adds complexity to the code and we're not sure if someone is using it anyway.

populating a carousel with a folder could be a nice shortcut at first, but I doubt people use it on a daily base; populating a carousel with a collection on the other side still makes some sense. any way, in none of our projects this is being used; all editors want to keep absolute control of the composite page.

fredvd: I guess that this change is because the rule for being addable to a container is no longer limited on contenttype, but only if it can provide a (lead) image and this image can also be providerd by folders,  so folders can no longer be unpacked? 

Just to avoid any misunderstanding My original question was only on the list tile expansion behaviour, for carousels I'm still fine if they're unpacked.  Alas there's not 'one use case to rule them all :-S'

hvelarde: yes, I know; but I stumbled upon the folder use case in the process of fixing that and I came into the conclusion it doesn't make sense neither because of what I explained above. also, by removing that behavior we are able to remove those useless adapters, simplifying the code even more.

fredvd: ok, that's the link, thanks for the clarification.

hvelarde:
@smcmahon I'll wait for your review until the end of the day; if no feedback is provided I'll merge, thanks!

src/collective/cover/interfaces.py
```diff
@@ -30,14 +32,8 @@ def getSearchResults(self, filter_portal_types, rooted, document_base_url, searc
...
-
-    def getUIDs(self):
-        """Get UUIDs associated with the object.
-            could be the UUID of the object or a
-            list of related UUIDs.
-
-        @return: iterable of UUIDs
-        """
+    """GenericUIDsProvider interface will be removed in collective.cover v1.7."""
+    warnings.warn(__doc__, DeprecationWarning)
```
idgserpro: We still have some doubts about how this deprecation should work.

Usually, deprecation warnings still keep the functionality and, in the specified version in the message warning, is removed. The [`replace_with_objects` method in collective.cover list tile in version 1.1b1 was a good example.](https://github.com/collective/collective.cover/blob/1.1b1/src/collective/cover/tiles/list.py#L223)

The way it's now, the warning seems useless, because it'll be shown in an already broken package. For example, we have a package that uses collective.cover as a dependency that uses this class and these adapters, when trying to upgrade to collective.cover 1.5b1 we get:

```
uuids = ICoverUIDsProvider(obj).getUIDs()
TypeError: ('Could not adapt', <ATImage at /plone/my-image>, <InterfaceClass collective.cover.interfaces.ICoverUIDsProvider>)
```

https://travis-ci.org/plonegovbr/brasil.gov.tiles/jobs/245449596#L1565

Any idea how we should solve this problem? Shouldn't the removal of these classes [(and adapters)](https://github.com/collective/collective.cover/pull/714/files#diff-2b38db85530a6440d8600df43b0f963bL88) be reverted (we can do a PR, we're not so sure exactly where to put the `warnings`)? If I add them "by hand" in a hacky manner in my eggs folder, the errors disappear.

idgserpro: Don't know if you're receiving this, so, FYI @hvelarde

hvelarde: IMO, the problem is not the code, but the missing ZCML registration; could you please test that?

idgserpro: The error is indeed because of the ZCML registration, but we need the provided funcionality: changing from "Could not adapt" to uuids always being `None` doesn't change much, since I could just do a `try: except TypeError`, set uuids to None and don't bother you in this PR anymore.

As we said, we don't mind removing these adapters, but they should be removed in 1.7 as is shown at the `warnings` message. We think that the deprecation implementation the way it is right now isn't right.

hvelarde: feel free to provide a solution on another PR; I asked for revision because of that, I don't want to break somebody else code but I have no time to test for all the implications.

idgserpro: No problem, we know that it's not possible to check for every possible corner case. If you agree with the PR do you think it's possible to make another release right after it's merging?

hvelarde: just a question: wouldn't be easier just to remove the requirement of the adapter from brasil.gov.tiles package before its new release?

idgserpro: No since it's a behavior the user is expecting when adding a container to a list. We can add the adapter to brasil.gov.tiles though, but we think the "deprecation" that was made here is wrong and that's why we're proposing to fix it here.

idgserpro: Before continuing, we did two really simple PR's that could be merged: https://github.com/collective/collective.cover/pull/736 and https://github.com/collective/collective.cover/pull/735.

Following the adapter issue, we've created https://github.com/collective/collective.cover/pull/737. If you agree with it, we need it merged and a collective.cover and a brasil.gov.tiles release, since we're upgrading collective.cover in IDGB and changes will be needed in brasil.gov.tiles after these both releases. We're planning on finishing the upgrade of collective.cover to 1.5b1 in IDGB tomorrow.

hvelarde: I'm on a training right now; I'll try to make the release but I can't compromise on having it ready early.

Issue #719, opened 2017-05-30
Tile edit overlay shows up twice with Plone 4.3-latest and p.a.widgets-1.x installed


petschki:
With latest Plone 4.3 and p.a.widgets 1.x branch installed the edit tile overlays show up twice. This is due to classname "pat-plone-modal" on a.edit-tile-link which registers the plone modal additionally to the $.fn.prepOverlay() in compose.js ... looks like this:

<img width="1035" alt="bildschirmfoto 2017-05-30 um 08 52 43" src="https://cloud.githubusercontent.com/assets/511761/26571363/ea3685ac-4515-11e7-8e31-2e5a59a7b990.png">

I don't want to disable prepOverlay because its used systemwide in Plone 4.3 .. my recommendation is to remove the pat-plone-modal classname as the Tiny editor for example isn't initialized correctly anyway. Without the pat-plone-modal class it looks OK:

<img width="1092" alt="bildschirmfoto 2017-05-30 um 08 59 44" src="https://cloud.githubusercontent.com/assets/511761/26571466/5c5e9aa2-4516-11e7-9d11-150a3b7fe1d8.png">

If you're with me, I'll make a small pull request on this.

petschki:
Well ... after some more investigations on the compose.js code I've changed my mind ;) I think it would be better to check for installed require.js **before** initalizing prepOverlay() and let the pat-plone-modal do it's job **instead** of prepOverlay() ... I've already fixed the JS code but right now I'm fighting with get_tinymce_options() to get the right context to adapt... maybe someone had this fight already?

petschki:
Well well ... time goes by ... I have to admit, that this issue sounded quite familiar. Looking at issue #543 I've realized, that I've already forked cover and implemented a working richt text modal. Now I've made a pull request which fixes Plone 4.3/p.a.widgets 1.x issues ... see #720

hvelarde:
rereading the issue I want to ask you if this happens with latest version of Plone only; if that the case then this needs to be fixed differently as should be related with https://github.com/plone/Products.CMFPlone/pull/1710.

let us know.

petschki:
I'm not sure what your mentioned issue has to do with this one ... but I can confirm, that this bug also occurs on Plone 4.3.11 with p.a.widgets 1.x ... same effect: double overlay.

hvelarde:
ok, then is not related; thanks!

Mubra:
Hi @petschki , I can't explain this behavior, will you have a recipe?

PR #720, opened 2017-05-30
Fix Plone 4.3 with p.a.widgets 1.x edit modals

petschki:
see issue #719

### DIFF
--- a/src/collective/cover/static/js/compose.js
+++ b/src/collective/cover/static/js/compose.js
@@ -96,7 +96,9 @@ $(document).ready(function() {
 
   TitleMarkupSetup();
 
-  if ($.fn.prepOverlay !== undefined) {
+  // if mockup is installed let pat-plone-modal do its job
+  // XXX: the check for require === 'undefined' should probably be more explicit
+  if ($.fn.prepOverlay !== undefined && typeof require === 'undefined') {
     $('a.edit-tile-link').prepOverlay({
       subtype: 'ajax',
       filter: '.tile-content',
@@ -140,26 +142,7 @@ $(document).ready(function() {
       },
       config: {
         onLoad: function() {
-          // With plone.app.widgets and Plone 4.3
-          if (typeof require !== 'undefined' && require.defined('pat-registry')) {
-            // Remove old editors references to work with ajax
-            if (typeof tinyMCE !== 'undefined' && tinyMCE !== null) {
-              if (tinyMCE.EditorManager != null) {
-                tinyMCE.EditorManager.editors = [];
-              }
-            }
-            // Add tinymce
-            $('.overlay textarea.mce_editable').addClass('pat-tinymce');
-            require('pat-registry').scan($('.overlay'), ['tinymce']);
-            // Wire save buttom to save tinymce
-            $( '.overlay input#buttons-save').on('click', function() {
-              tinyMCE.triggerSave();
-            });
-            // Hack to make overlay work over overlay
-            $('.overlay').on('mouseover', function() {
-              $('div.plone-modal-wrapper').css('z-index', '10050');
-            });
-          } else if (typeof initTinyMCE !== 'undefined') { // Plone 4.3
+          if (typeof initTinyMCE !== 'undefined') { // Plone 4.3
             // Remove old editors references to work with ajax
             if (typeof tinyMCE !== 'undefined' && tinyMCE !== null) {
               if (tinyMCE.EditorManager != null) {
--- a/src/collective/cover/widgets/richtext.py
+++ b/src/collective/cover/widgets/richtext.py
@@ -1,0 +1,50 @@
+import pkg_resources
+
+try:
+    pkg_resources.get_distribution('plone.app.widgets')
+except pkg_resources.DistributionNotFound:
+    pass
+else:
+    from Acquisition import aq_parent
+    from Products.CMFCore.utils import getToolByName
+    from collective.cover.tiles.richtext import IRichTextTile
+    from plone.app.widgets.base import dict_merge
+    from plone.app.widgets.dx import RichTextWidget
+    from plone.app.widgets.utils import get_tinymce_options
+    from z3c.form.interfaces import IFieldWidget, IFormLayer
+    from z3c.form.util import getSpecification
+    from z3c.form.widget import FieldWidget
+    from zope.component import adapter
+    from zope.interface import implementer
+
+
+    class TileRichTextWidget(RichTextWidget):
+        """ """
+
+        def _base_args(self):
+            args = {
+                'pattern': self.pattern,
+                'pattern_options': self.pattern_options.copy(),
+            }
+            context = aq_parent(self.wrapped_context())
+            args['name'] = self.name
+            properties = getToolByName(context, 'portal_properties')
+            charset = properties.site_properties.getProperty(
+                'default_charset', 'utf-8')
+            value = self.value and self.value.raw_encoded or ''
+            args['value'] = (self.request.get(
+                self.field.getName(), value)).decode(charset)
+
+            args.setdefault('pattern_options', {})
+            merged = dict_merge(
+                get_tinymce_options(context, self.field, self.request),
+                args['pattern_options'])
+            args['pattern_options'] = merged
+
+            return args
+
+
+    @adapter(getSpecification(IRichTextTile['text']), IFormLayer)
+    @implementer(IFieldWidget)
+    def TileRichTextFieldWidget(field, request):
+        return FieldWidget(field, TileRichTextWidget(request))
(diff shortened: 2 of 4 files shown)


petschki:
bah ... sorry for the mess in here ... tried to squash some commits, but gave up ... nevermind.

src/collective/cover/static/js/compose.js
```diff
@@ -96,7 +96,9 @@ $(document).ready(function() {
 
   TitleMarkupSetup();
 
-  if ($.fn.prepOverlay !== undefined) {
```
hvelarde: changes on JS usually require an upgrade step to cook the resources; do you mind to check if we already have one in place?

src/collective/cover/widgets/richtext.py
```diff
@@ -0,0 +1,50 @@
+import pkg_resources
+
+try:
```
hvelarde: I think this is not necessary as this code will only be execute if plone.app.widgets is installed and you already are checking for that on `configure.zcml`.

also, for me is pretty important to document what you're doing here, so we can remove this in the future when we drop compatibility with Plone 4. that sais, please add docstrings at the module and class levels explaining why this is needed..

petschki: sure .. you're absolutely right ... i'll remove that

hvelarde: we're still missing the docstrings.

petschki:
I'll try to fix the tests today ...

petschki:
@hvelarde everything is green except one test in DEXTERITY_ONLY setup ... and I really do not know whats going wrong there. Can you give me some pointers? Could I add this test to the noncritical list (like `issue_637` already is)?

src/collective/cover/static/js/compose.js
petschki: done that

hvelarde:
@petschki those are mandelbugs; I restarted the builds for you; what happened with the idea of working on this directly in plone.app.widgets?

petschki:
@hvelarde  ah ... thanks for restarting. Now we're green. But you're right about RichTextWidget... let me check p.a.widget and the pull request https://github.com/plone/plone.app.widgets/pull/160 before merging this...

petschki:
WIP: https://github.com/plone/plone.app.widgets/pull/163

petschki:
https://github.com/plone/plone.app.widgets/pull/163 was merged right now ... I'll request a release of p.a.widgets and cleanup the unused code here.

hvelarde:
do we still to add something in collective.cover, or just using the new widgets solve your problem?

petschki:
@hvelarde unfortunately we still have to map a custom FieldWidget adapter for the RichTextField (like its done in plone.app.widgets.dx_bbb) ... see https://github.com/plone/plone.app.widgets/issues/164 ... my latest pull request would fix this, so everything would work supermagically https://github.com/plone/plone.app.widgets/pull/165

hvelarde:
ok, just keep me updated :-)

hvelarde:
@petschki do you mind to make a rebase? also, do you mind to update me the status on this after the changes you made on plone.app.widgets?

petschki:
@hvelarde sure ... I'll rebase later today. the `plone.app.widgets` progress made the integration a bit easier, so we do not need to wire any widgets in collective.cover anymore. right now, it depends on the latest `1.x` branch but I think we can have a p.a.widgets release 1.10 this week. I've also setup a testing environment for the widgets implementation but got stuck in robots test, not finding elements within the plone-modal ... need to do more investigations on that this week. I'll commit the `4.3.x-widgets` travis-environment later today and put it to allow_failures ...

hvelarde:
we're having some mandelbugs with RF tests right now and I need to restart failing builds from time to time.

I think the problem on the new build you just created seems to be related with a feature removed from zc.buildout, but I don't remember very well and I'm not sure.

FWIW, buildout previously was able to downgrade itself and I think that's not the case anymore; that's why you're having:

```python
VersionConflict: (zc.buildout 2.9.4 (/home/travis/build/collective/collective.cover/eggs/zc.buildout-2.9.4-py2.7.egg), Requirement.parse('zc.buildout==1.7.1'))
```

probably @idgserpro knows what is happening here.

idgserpro:
We have good news and bad news. Let's start with the good ones, the reason (and a possible workaround) of the error.

When you use bootstrap.py, it downloads the latest zc.buildout (that now is 2.9.4). In `bin/buildout` it will point to `/home/travis/build/collective/collective.cover/eggs/zc.buildout-2.9.4-py2.7.egg`.

Here is the insanity: after bootstraping, when running `bin/buildout`, when **downloading** and extracting zc.buildout 1.7.1 for the *first* time when it's a direct dependency (`install_requires`) in an extensions section, it will give you the error: but if you already have a `/home/travis/build/collective/collective.cover/eggs/zc.buildout-1.7.1-py2.7.egg` in place, the buildout continues. That's why the error was happening only in the new job with the mr.developer in extensions, the old ones don't have mr.developer and had a cache that the new job didn't.

You can simulate it in your machine instead of travis, in a clean virtualenv environment (remember to comment a `eggs-directory` in `~/.buildout/default.cfg` if you have it in place). Funny thing is: the first time you run it will break (giving the error), but the second time won't, because you now have a zc.buildout-1.7.1-py2.7.egg in your environment.

The bad news: It seems that it started in zc.buildout 2.9.x branch (https://travis-ci.org/collective/collective.cover/builds/247259422) but we don't have time to check exactly what happened to start having this issue. As miohtama used to say, *life is hard and Plone is harder™*.

`setuptools` have the same issue as zc.buildout above: travis decides to [automatically install some packages](https://docs.travis-ci.com/user/languages/python/#Pre-installed-packages) and `setuptools` is a direct dependency from some of them (the specific versions can be [checked here](https://travis-ci.org/collective/collective.cover/builds/247295792#L150)).

There are two immediate workarounds:

- Pin zc.buildout and setuptools to the last version in versions-4.3.x-widgets.cfg (https://travis-ci.org/collective/collective.cover/jobs/247501030);
```
# VersionConflict: (zc.buildout 2.9.4 (/home/travis/build/collective/collective.cover/eggs/zc.buildout-2.9.4-py2.7.egg), Requirement.parse('zc.buildout==1.7.1'))
# We still don't know why this happens. It seems that when using extensions with a package that
# requires zc.buildout (like mr.developer).
zc.buildout = 
# XXX: Same as above.
# VersionConflict: (setuptools 12.0.5 (/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages), Requirement.parse('setuptools==20.1.1'))
# travis has some packages installed by default in it's virtualenv and setuptools is installed.
# https://docs.travis-ci.com/user/languages/python/#Pre-installed-packages
setuptools =
```
- Run bootstrap with Plone 4.3 pins of zc.buildout and setuptools, `bootstrap.py --buildout-version=1.7.1 --setuptools-version=20.0.1` in `.travis.yml` (https://travis-ci.org/collective/collective.cover/jobs/247520681)
```
# XXX:
# Trying to pin in versions-4.3.x-widgets to the latest version it works but then all tests break.
# So we're going to use the same versions that Plone uses in it's versions.
- PLONE_VERSION=4.3 WIDGETS=true BOOTSTRAP_PARAMS="--buildout-version=1.7.1 --setuptools-version=20.1.1"
(...)
python bootstrap.py $BOOTSTRAP_PARAMS
```

Bad news part 2: now you have **20** failed tests after solving the `VersionConflict`. @petschki, can you try to run locally and see if this happens only in Travis?

Another question: why are you pinning `plone.app.widgets = ` since you're already checking it's branch in `sources`?

petschki:
@idgserpro thanks for the details ... pinned everything down and removed unneeded version pin of plone.app.widgets. not the buildout runs smooth ... I'm aware of those 20 errors. thats why I moved the environment to allow_failures ..

idgserpro:
You don't need the two pins, only one of them. It's a good practice as well to document the pins in the cfg itself, so in the future if you try to remove them its less work to know why it was pinned. Our suggestion to doc above the pins:

```
# XXX: https://github.com/collective/collective.cover/pull/720#issuecomment-311383938
```

hvelarde:
I not sure which one would be best practice here; I guess we should ask.

idgserpro:
I think we should open an issue at zc.buildout exposing the problem and, from there, we can know which one is the best approach.

From a Plone perspective, since they are planning to get rid of bootstrap.py, they will problably be in favour of changing cfgs.

src/collective/cover/static/js/compose.js
```diff
@@ -96,8 +96,18 @@ $(document).ready(function() {
...
-    $('a.edit-tile-link').prepOverlay({
+  $('body').on('click', 'a.edit-tile-link', function(e) {
+
+    // XXX: need to check for body class here, because it might not been when document.ready() :(
+    if($("body").hasClass("pat-plone-widgets")) {
+      // XXX: reload tile content asynchronously here when destroying the modal
+      //      need to find the right event where to hook in
+      return  // exit here
```
rodfersou: miss semicolon here

ale-rt (approved):

hvelarde (changes requested):
the branch is outdated; a review is not possible in this state.

PR #725, opened 2017-06-13
Drop support for Plone 4.2

hvelarde:


idgserpro:
Remove [this](https://github.com/collective/collective.cover/pull/725/files#diff-354f30a63fb0907d4ad57269548329e3R56).

hvelarde:
@rodfersou @idgserpro done!

fredvd:
Makes sense to drop support for 4.2. only supported 4.X version is 4.3. If organisations really want to continue using Plone they should upgrade to 4.3 and if they don't care about security releases then the last release can still be used on Plone 4.2.

PR #730, opened 2017-06-20
Improve doc of collective.js.bootstrap pin.

idgserpro:
[ci skip]

### DIFF
--- a/versions-4.3.x.cfg
+++ b/versions-4.3.x.cfg
@@ -14,6 +14,7 @@ test-eggs =
     plone.app.referenceablebehavior
 
 [versions]
+# XXX: https://github.com/collective/collective.cover/issues/492
 collective.js.bootstrap = 2.3.1.1
 plone.app.blocks = 4.0.6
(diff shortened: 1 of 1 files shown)


versions-4.3.x.cfg
```diff
@@ -14,6 +14,7 @@ test-eggs =
     plone.app.referenceablebehavior
 
 [versions]
+# https://github.com/collective/collective.cover/issues/492
```
hvelarde: add `XXX: ` in front to keep consistency.

idgserpro: Done

idgserpro:
I think we have a problem in github's workflow for accepting a PR when a `[ci skip]` is added to a commit. @mauritsvanrees have you ever seen something like this?

idgserpro:
Just out of curiosity @hvelarde, is there a controlpanel where you can "retreat" the pre-requisites of a PR approval?

hvelarde:
as an administrator I can skip those requirements :sunglasses:

PR #732, opened 2017-06-20
Run RF tests on PhantomJS instead of Firefox

hvelarde:


hvelarde:
@datakurre could you please give some advice here? do you know why tests are failing?

idgserpro:
Shouldn't we test on both?

datakurre:
I guess this is because Firefox selenium dependencies are currently really hard, because of each new versions requiring too often both the latest selenium and the latest geckodriver.

@hvelarde PhantomJS may run in mobile window size by default. Do you already have ``Set window size  1200  900`` or similar in the tests?

hvelarde:
@datakurre no, I don't; that's the recommended size? I'll test later.

hvelarde:
@datakurre seems it's not working neither; let me know if you have another idea.

idgserpro:
Why phantomjs instead of chrome? Do you want to get rid of drivers nightmare?

hvelarde:
@idgserpro yes

hvelarde:
superseded by #784

PR #735, opened 2017-06-22
Use the same plone.app.jquery that Plone uses.

idgserpro:
Instead of 1.7.2, 1.7.2.1.

http://dist.plone.org/release/4.3.14/versions.cfg

http://dist.plone.org/release/4.3-latest/versions.cfg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
   matrix:
     - PLONE_VERSION=4.3
     - PLONE_VERSION=4.3 DEXTERITY_ONLY=true
-    - PLONE_VERSION=4.3 JQUERY=1.9.1.3
+    - PLONE_VERSION=4.3 VERSIONS=versions:plone.app.jquery=1.9.2
     - PLONE_VERSION=5.0
     - PLONE_VERSION=5.1
 matrix:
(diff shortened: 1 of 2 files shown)


hvelarde:
there's a better way for doing this.

.travis.yml
```diff
@@ -17,18 +17,17 @@ env:
   matrix:
     - PLONE_VERSION=4.3
     - PLONE_VERSION=4.3 DEXTERITY_ONLY=true
-    - PLONE_VERSION=4.3 JQUERY=1.9.1.3
+    - PLONE_VERSION=4.3 VERSIONS=versions:plone.app.jquery=1.9.2
```
idgserpro: Do you have any documentation about the `versions:` trick in buildout?

hvelarde: you can pass any part/directive value on command line to override those on configuration; that's on buildout documentation:

https://travis-ci.org/collective/collective.cover/jobs/245957084#L1805

idgserpro: This is **extremely** useful in testing, instead of changing buildout.cfg everytime you need to play with different versions. Nice find!

idgserpro: Indeed, it's in official docs, first paragraph :laughing: : http://docs.buildout.org/en/latest/reference.html

idgserpro: @hvelarde Really nice, `bin/buildout buildout:eggs+=plone.app.drafts` for example works as well. This is a really nice trick for testing and CI. For example, instead of using `DEXTERITY_ONLY=true` and trying to read an environment variable, you could have `bin/buildout buildout:eggs+=plone.app.contenttypes` in a different job.

hvelarde: yes, that's on my radar; I've done that in other packages already.

Issue #718, opened 2017-05-28
Uncaught TypeError: $(...).tabs is not a function


zopyx:
Latest Plone 5.1 + collective.cover 1.4b1 gives the following error as soon as I click on "Compose"

plone-legacy-compiled.js:19 Uncaught TypeError: $(...).tabs is not a function
    at contentchooserMaker (plone-legacy-compiled.js:19)
    at HTMLDocument.<anonymous> (plone-legacy-compiled.js:19)
    at j (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at Function.ready (jquery.min.js:2)
    at HTMLDocument.J (jquery.min.js:2)

zopyx:
The problem is independent of the following version pins:
```

[versions]
collective.js.jqueryui = 1.8.16.9
plone.app.jquery = 1.7.2
plone.app.jquerytools = 1.5.7
plone.app.z3cform = 0.6.3
plone.directives.form = 1.1
```

hvelarde:
this package is still not compatible with Plone 5; I'm adding this issue to the list of things that need to be fixed, but I have to warn you that I have still no interest on fix this as we don't have a use case and I think we should wait until the RR is fixed.

rodfersou:
here we can use as base this referenced PR, but keep the tabs function just for plone 4.x

PR #744, opened 2017-08-11
use pat-autotoc instead of old formTabbing refs #718

cdw9:
takes care of the JS `tabs()` error, which break most JS on pages for anonymous users.

### DIFF
--- a/src/collective/cover/browser/templates/content_contentchooser.pt
+++ b/src/collective/cover/browser/templates/content_contentchooser.pt
@@ -7,17 +7,14 @@
       tal:omit-tag="">
 <body tal:omit-tag="">
 
-<div id="contentchooser-content-search" class="right ui-widget-content">
+<div id="contentchooser-content-search" class="right ui-widget-content pat-autotoc"
+     data-pat-autotoc="levels: legend; section: fieldset; className: autotabs">
     <div class="close">
         <a href="#" class="hiddenStructure" title="Close this box">Close this box.</a>
     </div>
-    <ul class="formTabs">
-      <li class="formTab"><a href="#recent" i18n:translate="">Recent items</a></li>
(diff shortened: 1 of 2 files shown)


idgserpro:
It would be a nice touch to run `rebuild_i18n` since you're adding new strings to the package. `Recent Items` is `Itens recentes` in pt-br, @hvelarde could help with Spanish `*.po`.

idgserpro:
Since js files were modified, an upgradeStep is needed as well (for example https://github.com/collective/collective.cover/blob/master/src/collective/cover/upgrades/v18/configure.zcml#L15).

hvelarde:
@cdw9 thanks! is this for Plone 5 only? did you test if it works also on Plone 4.3?

cdw9:
@hvelarde Sorry, I did not test on 4.3, I was assuming master would be for Plone 5.  This will not work on 4.3. I see there is a plone5 branch, but it hasn't been touched in a couple years. Should I get that updated and use it instead?

@idgserpro There are not new strings, they were just moved, but I missed moving the `i18n:translate=""` with them. I can get that added. Also we won't need to cook the registries since this is only for Plone 5.

hvelarde:
@cdw9 you may base your work in the master branch as we need to keep moving slowly for now.

we don't use collective.cover in Plone 5 and having a versions specific to it will increase the administrative work.

cdw9:
closing PR: I will fix a couple things and resubmit to the `plone5` branch

hvelarde:
@cdw9 please don't use that branch as it's completely outdated and I already implemented many of those things in the master branch.

it's really better to fix things in master as we can't compromise ourselves to maintain 2 branches of this package: we are a very small team with a lot of work to do; please help us on this.

PR #746, opened 2017-08-28
Issue 745 fix idatetimewidget if ploneappcontenttypes

idgserpro:
A test was added to make sure the customization works. Just for the sake of testing, removing the fix in https://github.com/collective/collective.cover/compare/issue_745_fix_idatetimewidget_if_ploneappcontenttypes?expand=1#diff-5cb82c667fc9951b7e74bc556bd9e6f2R41 and leaving just the tests:

No plone.app.contenttypes (green):
https://travis-ci.org/collective/collective.cover/jobs/269280818

With plone.app.contenttypes (breaks, as expected: `AssertionError: Element 'css=select#collective-cover-basic-date-format option[value=datetime]' did not appear in 30 seconds`):
https://travis-ci.org/collective/collective.cover/jobs/269280819

After merging this, a new release would be nice, since this error was found while brasil.gov.portal 1.2b1 was being tested.

Thanks in advance!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,12 @@ There's a frood who really knows where his towel is.
 1.6b2 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- Fix IDatetimeWidget tile override if using plone.app.contenttypes >= 1.1.1:
+  collective.z3cform.datetimewidget is merged into plone.formwidget.datetime,
+  so the zcml must override the template from plone.formwidget.datetime.z3cform.interfaces.IDatetimeWidget
+  as well. (closes `#745`_).
+  [idgserpro]
+
 - Review tile refresh using custom event.
   [rodfersou]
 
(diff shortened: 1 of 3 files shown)


hvelarde:
why don't we depend only on `plone.formwidget.datetime` instead?

idgserpro:
> why don't we depend only on plone.formwidget.datetime instead?

More tests would be needed, possibly an upgradeStep (to uninstall collective.z3cform.datetimewidget, or at least a README stating the need to uninstall it) and we don't know if it's a drop in replacement with no caveats for collective.z3cform.datetimewidget. We thought about a pragmatic approach like the one in this PR (keeping both registrations in zcml) would be a better fit.

plone.formwidget.datetime was added to plone.app.event but *removed* as well somewhere in time (https://github.com/plone/plone.app.event/commit/c72829947d18a21797921dac2bf6f0fbcad25088) but we didn't have time to check this whole widget ecosystem, it's changes and motivations: that's why the PR was simple and mainly to fix a regression bug.

We think the whole widgets customization improvements can be adressed in another PR.

idgserpro:
github and Travis mangled all builds references because we created a branch of the branch to show the error in https://github.com/collective/collective.cover/pull/746#issue-253405060

The PR is green, check https://travis-ci.org/collective/collective.cover/builds/269289164.

hvelarde:
I didn't understand the issue; please open a new PR so we can merge safely.

idgserpro:
This PR has become too cluttered. Check https://github.com/collective/collective.cover/pull/748

Issue #745, opened 2017-08-28
IDatetimeWidget customization is gone if plone.app.contenttypes >= 1.1.1 is installed


idgserpro:
In a basic tile, this is what you get when configuring it in Layout tab:

![No plone.app.contenttypes 1.1.1](https://user-images.githubusercontent.com/8203476/29781596-15c85968-8bf0-11e7-9b2f-25c978aa474d.png)

And this is what you get after installing plone.app.contenttypes 1.1.1:

![With plone.app.contenttypes 1.1.1](https://user-images.githubusercontent.com/8203476/29781602-188945fe-8bf0-11e7-9650-d8b5b20918ba.png)

This happens because at this plone.app.contenttypes version, it installs [plone.app.event](https://github.com/plone/plone.app.contenttypes/blob/03fb2d9cc7be4435d6c4596e231c998a8088dc61/plone/app/contenttypes/profiles/default/metadata.xml#L5), which installs [plone.formwidget.datetime](https://github.com/plone/plone.app.event/blob/1.1.6/plone/app/event/profiles/default/metadata.xml#L6), which is used instead of collective.z3cform.datetimewidget (which is customized in https://github.com/collective/collective.cover/blob/059fecac5688c33859700e0352aaa6d99971af05/src/collective/cover/tiles/configuration_widgets/configure.zcml).

A fix will be done in a PR.

Issue #747, opened 2017-08-28
Review IDatetimeWidget dependencies for customization


idgserpro:
If you don't have plone.app.contenttypes >= 1.1.1 installed, IDatetimeWidget comes from collective.z3cform.datetimewidget.

If you have plone.app.contenttypes >= 1.1.1 installed, [plone.app.event 1.1.6 is installed as well](https://github.com/plone/plone.app.contenttypes/blob/03fb2d9cc7be4435d6c4596e231c998a8088dc61/plone/app/contenttypes/profiles/default/metadata.xml#L5) and the IDatetimeWidget comes from plone.formwidget.datetime (because [plone.app.event 1.1.6 installs it](https://github.com/plone/plone.app.event/blob/1.1.6/plone/app/event/profiles/default/metadata.xml#L6)). 
> plone.formwidget.datetime is a merge of collective.z3cform.datetimewidget and archetypes.datetimewidget (which itself was derived from collective.z3cform.datetimewidget).

From plone.app.event > 1.2, [plone.formwidget.datetime is removed](https://github.com/plone/plone.app.event/commit/c72829947d18a21797921dac2bf6f0fbcad25088):

> dropped. The new plone.app.widgets package implements archetypes and z3cform

This issue was opened to discuss if it's possible to use just one widget dependency for all situations in the future, avoiding all this mess. Until this is done, multiple zcml registrations in https://github.com/collective/collective.cover/blob/059fecac5688c33859700e0352aaa6d99971af05/src/collective/cover/tiles/configuration_widgets/configure.zcml#L40 will be done.

PR #748, opened 2017-08-28
Issue 745 fix idatetimewidget if ploneappcontenttypes

idgserpro:
This fixes https://github.com/collective/collective.cover/issues/745, check the issue for more information. This whole collective.z3cform.datetimewidget and plone.formwidget.datetime dependency situation here can be improved in the future by discussing https://github.com/collective/collective.cover/issues/747 and deciding the best approach.

If possible, a new release after merge would be a good idea.


idgserpro:
Would you mind creating a new release so we can focus on finishing IDG 1.2b1? :)

PR #749, opened 2017-08-29
Update versions of dependencies being used

hvelarde:


README.rst
```diff
@@ -116,9 +116,9 @@ Edit your buildout.cfg and add add the following to it:
     [versions]
     ...
     collective.js.bootstrap = 2.3.1.1
-    plone.app.blocks = 4.0.6
+    plone.app.blocks = 4.1.0
```
idgserpro: Is there anyway that you can add a reference here in the docs to https://github.com/collective/collective.cover/blob/ae9eebdb1fe00b3f829a179992bd889df4ffd0e0/versions-4.3.x.cfg#L16 or would it be more complicated?

versions-4.3.x.cfg
```diff
@@ -16,9 +16,9 @@ test-eggs =
 [versions]
 # XXX: https://github.com/collective/collective.cover/issues/492
```
idgserpro: Maybe instead of doing what I said in the last command, put a `# Remember to edit README.rst if you change any dependency here`.

hvelarde: I think that's not necessary at this time.

PR #755, opened 2017-09-21
Miscellaneous fixes

hvelarde:


.travis.yml
```diff
@@ -36,14 +36,9 @@ script:
 - bin/code-analysis
 - bin/test
 after_success:
-- bin/createcoverage -t "--layer=!Robot"
+- bin/createcoverage --output-dir=htmlcov -t "--layer=!Robot"
```
idgserpro: I'm missing a doc about the usage of this parameter: it seems to be used just to be a workaround for Firefox instead of related to the implementation.

hvelarde: I connected the dots after @mauritsvanrees response to the other issue: if we don't use this parameter coverage runs and opens a browser window to show the report; the browser window left open was causing the original failure.

Issue #757, opened 2017-10-07
@@edit-tile/collective.cover.carousel view breaks when special character is in title of one carousel item


fulv:
Plone 4.3.15
collective.cover master branch (today)

1. Create an Image content item
2. Put a special character, e.g. ™ (trademark symbol) in the title of the Image item 
3. Create a cover
4. Add a carousel tile to the layout
5. Compose:  drag the Image into the carousel
6. View.  Everything is fine.
7. Compose:  click the "Edit" link of the carousel tile

Result:  The edit tile layover dialog start appearing, but stays empty and very small (see screenshot).
See traceback at the bottom.

I have conclusively traced this to the presence of a special character in the title.  If I edit the Image and remove the special character from the title, the issue disappears.

Since the ™ does not cause problems anywhere else, not even in the normal view of the carousel in the cover, I'm sure I can find an easy fix.  It'll probably be in `collective.cover.widgets.textlinessortable.TextLinesSortableWidget.get_custom_title` or in the `textlines_sortable_input.pt` template.  The `get_custom_description` method probably has the same issue.

![screenshot 2017-10-07 at 10 39 07](https://user-images.githubusercontent.com/460742/31306304-3f6e18bc-ab4d-11e7-9e19-ec4606e6d6ca.png)

```
2017-10-07 10:43:37 ERROR Zope.SiteErrorLog 1507365817.190.86532871068 http://localhost:8080/Plone/cover/@@edit-tile/collective.cover.carousel/fd9b43b7-c6a8-4647-a62d-84758abe5850
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module collective.cover.tiles.edit, line 110, in __call__
  Module plone.z3cform.layout, line 66, in __call__
  Module plone.z3cform.layout, line 60, in update
  Module plone.app.tiles.browser.base, line 78, in render
  Module z3c.form.form, line 162, in render
  Module zope.browserpage.viewpagetemplatefile, line 51, in __call__
  Module zope.pagetemplate.pagetemplate, line 132, in pt_render
  Module zope.pagetemplate.pagetemplate, line 240, in __call__
  Module zope.tal.talinterpreter, line 271, in __call__
  Module zope.tal.talinterpreter, line 343, in interpret
  Module zope.tal.talinterpreter, line 888, in do_useMacro
[...]

PR #758, opened 2017-10-07
Fixes issue #757

fulv:


src/collective/cover/profiles/testfixture/structure/.objects
```diff
@@ -1,3 +1,4 @@
 my-image,Image
 my-image1,Image
 my-image2,Image
+my-image3,Image
```
djowett: I'm not sure you need this extra image - and it seems to be breaking the robot tests a bit more than usual!
https://travis-ci.org/collective/collective.cover/jobs/284702469

src/collective/cover/widgets/textlinessortable.py
```diff
@@ -80,7 +80,7 @@ def get_custom_title(self, uuid):
             return title
         # If didn't find, get object title
         obj = uuidToObject(uuid)
-        return obj.Title()
+        return obj.Title().decode('utf-8')
```
djowett: is using `safe_unicode()` not, well, safer?  I think (but not sure) there's a danger that obj.Title() may be e.g. u'\xc6\xb5' (as opposed to '\xc6\xb5') in which case using `decode('utf-8')` will break.

src/collective/cover/widgets/textlinessortable.py
```diff
@@ -99,7 +99,7 @@ def get_custom_description(self, uuid):
             return description
         # If didn't find, get object description
         obj = uuidToObject(uuid)
-        return obj.Description()
+        return obj.Description().decode('utf-8')
```
djowett: Again, question if using safe_unicode() is safer

src/collective/cover/tests/test_textlinessortable_widget.py
```diff
@@ -158,20 +158,65 @@ def test_extract(self):
         self.assertDictEqual(extracted_value, expected)
 
     def test_utf8_custom_data(self):
-        obj = self.portal['my-image']
-        obj.setDescription('áéíóú')
+        obj1 = self.portal['my-image']
+        obj1.setDescription('áéíóú')
+        obj2 = self.portal['my-image3']
```
djowett: IIUC my-image2 will do, no?

src/collective/cover/tests/test_textlinessortable_widget.py
```diff
@@ -158,20 +158,65 @@ def test_extract(self):
         self.assertDictEqual(extracted_value, expected)
 
     def test_utf8_custom_data(self):
```
djowett: I was seeing the same error without customising the title, 
(in fact I found it difficult to reproduce when there WAS a custom title)
So for example, the following gives an error for me:

        title = u'El veloz murciélago hindú comía feliz cardillo y kiwi'
        obj = self.portal['my-image']
        obj.setTitle(title)
        obj.reindexObject()
        self.tile.populate_with_object(obj)

        self.assertIsInstance(self.tile.get_title(obj), unicode)
        self.assertEqual(self.tile.get_title(obj), title)

maybe this supplements what you've already done?

src/collective/cover/tests/test_textlinessortable_widget.py
fulv: IIUC, whether you use the title of the image object or set a custom title you will be hitting different code paths, which is why the existing tests did not catch this issue.  That is why I created a `my-image3` which has both a title and description containing accented letters.
I don't understand the robot test failures, though.
Maybe I'll try it without `my-image3` first to see if the failures go away.

src/collective/cover/widgets/textlinessortable.py
fulv: Yes, thanks, I'll change that.   I had a nagging suspicion.

fulv:
Weird, that seems to have fixed the robot tests.  I have no idea why the robot tests depend on the `structure` GenericSetup test profile.

djowett:
@fulv, It seems to be failing in the collection tile test and I wonder if it's something to do with the default ordering of the collection - which I don't know, but could be modification date & since you don't add a corresponding line [here](https://github.com/collective/collective.cover/blob/44da774bbd799a527a0fdb36c6983ef3d5eb8e0c/src/collective/cover/testing.py#L169) for image-3 - perhaps that becomes the "oldest" image in the collection, and therefore the first item matching the `.tile-header h1` selector [in the failing line of the robot test](https://github.com/collective/collective.cover/blob/44da774bbd799a527a0fdb36c6983ef3d5eb8e0c/src/collective/cover/tests/test_collection_tile.robot#L82)

Just a wild guess, HTH.  I'll approve the changes though

hvelarde:
@fulv I need to make a release soon to include latest changes I made on the package yesterday; would you mind to fix this so we can merge it before?

fulv:
Yes, I'll fix tomorrow.

djowett:
Thanks guys! Good job!

fulv:
thank you!

hvelarde:
thanks to you both; release 1.6b3 is out since yesterday: https://pypi.python.org/pypi/collective.cover/1.6b3

PR #760, opened 2017-10-27
use pat-autotoc instead of old formTabbing refs #718

abosio:
@hvelarde This is the fix I was referring to.


hvelarde:
awesome! we need to fix the related tests so we can merge this feature.

rodfersou:
we did almost the same here https://github.com/collective/collective.cover/pull/791/files already in master branch, thank you

PR #762, opened 2017-10-30
Revert declaring cover_layout field in content type schema as readonly

### LINKED ISSUE #761: TypeError: Can't set values on read-only fields (name=cover_layout, class=collective.cover.content.Cover)
Seems the workaround introduced in #700 was in fact a bad idea and is causing other issues like the one descibed in https://github.com/plonegovbr/brasil.gov.portal/issues/388

We need to find another way to fix this, probably the one described in https://github.com/plone/plone.app.testing/issues/39

hvelarde:
fixes #761 
refs: https://github.com/plone/plone.app.testing/issues/39

### DIFF
--- a/src/collective/cover/testing.py
+++ b/src/collective/cover/testing.py
@@ -116,11 +116,13 @@ def generate_jpeg(width, height):
 
 
 # FIXME: workaround for https://github.com/plone/plone.app.testing/issues/39
-autoform = ('plone.autoform', {'loadZCML': True})
-tinymce = ('Products.TinyMCE', {'loadZCML': True})
-products = list(PLONE_FIXTURE.products)
-products.insert(products.index(tinymce), autoform)
-PLONE_FIXTURE.products = tuple(products)
+#        Products.TinyMCE is used only in Plone 4
+if not IS_PLONE_5:
+    autoform = ('plone.autoform', {'loadZCML': True})
+    tinymce = ('Products.TinyMCE', {'loadZCML': True})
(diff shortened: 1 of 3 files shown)


src/collective/cover/testing.py
```diff
@@ -115,6 +115,16 @@ def generate_jpeg(width, height):
...
 
 
+# FIXME: workaround for https://github.com/plone/plone.app.testing/issues/39
+#        Products.TinyMCE is used only in Plone 4
+if not IS_PLONE_5:
+    autoform = ('plone.autoform', {'loadZCML': True})
+    tinymce = ('Products.TinyMCE', {'loadZCML': True})
+    products = list(PLONE_FIXTURE.products)
+    products.insert(products.index(tinymce), autoform)
+    PLONE_FIXTURE.products = tuple(products)
```
idgserpro: We have to do this in ```brasil.gov.portal``` if we want to update Plone.

But honestly I think this should be fixed in plone.app.testing.

hvelarde: I need to keep going on with development; we have deadlines.

idgserpro:
Then put it on ```brasil.gov.portal```.

src/collective/cover/testing.py
idgserpro: Then put it on ```brasil.gov.portal```.

hvelarde: done in d530a062cbc8b9fc71c0d38fcd1c2ed23da32825

PR #764, opened 2017-11-02
Do not auto include package dependencies, but declare them explicitly

hvelarde:


idgserpro:
Just to know how to review, what's the motivation of this?

src/collective/cover/configure.zcml
```diff
@@ -10,8 +10,14 @@
...
 
-  <includeDependencies package="." />
+  <include package="collective.js.bootstrap" />
+  <include package="collective.js.galleria" />
+  <include package="plone.app.blocks" />
+  <include package="plone.app.dexterity" />
+  <include package="plone.app.iterate" />
+  <include package="plone.app.lockingbehavior" />
   <include package="plone.app.relationfield" />
+  <include package="plone.app.tiles" />
```
rodfersou: Isn't it needed to include every dependency from [metadata.xml](https://github.com/collective/collective.cover/blob/master/src/collective/cover/profiles/default/metadata.xml#L5-L11)?

How about to both lists be equal?

hvelarde: including a package in `configure.zcml` does one thing and including it in `metadata.xml` does another; a package in `metadata.xml` should be included in `configure.zcml` but the opposite is not true.

PR #777, opened 2018-03-05
Provide alternative text for image fields in tiles

### LINKED ISSUE #628: Allow to edit the alt attribute on tiles including images
We must provide text alternatives for any non-text content so that it can be changed into other forms people need, such as large print, braille, speech, symbols or simpler language.

For more information see: https://www.w3.org/TR/UNDERSTANDING-WCAG20/text-equiv.html

hvelarde:
- [x] banner tile
- [x] basic tile
- [x] others? (other tiles just reference objects)

closes #628

### DIFF
--- a/src/collective/cover/tiles/banner.py
+++ b/src/collective/cover/tiles/banner.py
@@ -3,7 +3,9 @@ from Acquisition import aq_base
 from collective.cover import _
 from collective.cover.tiles.base import IPersistentCoverTile
 from collective.cover.tiles.base import PersistentCoverTile
+from collective.cover.tiles.configuration_view import IDefaultConfigureForm
 from collective.cover.utils import get_types_use_view_action_in_listings
+from plone.autoform import directives as form
 from plone.namedfile import field
 from plone.tiles.interfaces import ITileDataManager
 from plone.uuid.interfaces import IUUID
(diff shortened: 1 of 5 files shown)


src/collective/cover/tiles/banner.py
```diff
@@ -110,5 +125,6 @@ def htmltag(self):
 
     @property
     def alt(self):
-        """Return the alt attribute for the image."""
-        return self.data.get('description') or self.data.get('title')
+        """Return alternative text dealing with form init issues."""
+        alt_text = self.data['alt_text']
+        return alt_text if alt_text is not None else u''
```
idgserpro: Shouldn't this test `alt_text` and, if not available, get from self.data.get('description') or self.data.get('title')?

hvelarde: no, IMO; read this: https://webaim.org/techniques/alttext/

idgserpro: Imagine a website with thousands of covers, will I need to enter in each one and provide the `alt_text`? How do we handle this?

hvelarde: as per @agnogueira suggestion I'm going to initialize the field with the image description, but yes: this is one of the hardest parts of accessibility.

hvelarde: it's not that easy as in most of the cases we don't have an Image, but a content type with an image field. IMO, the behavior like I implemented it is the right way of doing it, but I don't want to have an endless discussion about it right now.

resuming: **using the title or description of a content type to describe its lead image is completely wrong and I'm going to change that to an empty string by default in collective.cover 2**; right now, I'm going to leave the current behaviour.

idgserpro: I didn't get it, you say you're going to leave the behavior as is but merged the different implementation? Are you going to do this in another PR?

hvelarde: this specific part of the code only takes care of the `alt` attribute when no object has been dragged into the tile (remember that default values on the schema don't work): the `alt_text` field is none and the attribute is removed from the tag on the template if I don't do this.

PR #779, opened 2018-04-24
Select multiple class in cover configuration

rodfersou:
WIP


src/collective/cover/browser/templates/layoutedit.pt
```diff
@@ -51,13 +51,9 @@
...
                     </div>
                     <div id="class-chooser" title="Choose Class">
-                        <select>
-                            <tal:classes repeat="item view/css_classes">
-                                <option tal:attributes="value item/value"
-                                        tal:content="item/title">
-                                </option>
-                            </tal:classes>
-                        </select>
+                        <input class="cssclasswidget" type="button" value="Select options ▼"
```
hvelarde: `value` needs to be internationalized.

src/collective/cover/static/css/layout_edit.css
```diff
@@ -731,3 +731,45 @@ input[type='submit'].btn::-moz-focus-inner
 {
     cursor: move;
 }
+body .cssclasswidget
```
hvelarde: how are we going to deal with intermediate caching servers and this change?

src/collective/cover/static/js/layout_edit.js
```diff
@@ -1,4 +1,112 @@
 (function($) {
+  var CSSClassWidget = (function() {
```
hvelarde: how are we going to deal with intermediate caching servers and this change?

src/collective/cover/tiles/configuration_widgets/cssclasswidget.pt
```diff
@@ -6,43 +6,10 @@
...
-             tal:condition="not:item/selected"
-             tal:attributes="id item/id;
-                             value item/value"
-             tal:content="item/content">label</option
-    ></tal:block>
-    </select>
-
-</html>
\ No newline at end of file
+    <input class="cssclasswidget" type="button" value="Select options ▼"
```
hvelarde: `value` needs to be internationalized.

src/collective/cover/tiles/configuration_widgets/cssclasswidget.py
```diff
@@ -21,6 +21,21 @@ def update(self):
         if isinstance(self.context.get('css_class'), unicode):
             self.value = [self.context.get('css_class')]
 
+    def options(self):
+        items = [
+            {
+                key: value
+                for key, value in item.iteritems()
```
hvelarde: `iteritems()` works on Python 2 only; you need to import it from `six`; check http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items

CHANGES.rst
```diff
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.6b6 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- Select multiple CSS class in cover configuration.
```
hvelarde: "Allow selection of multiple CSS classes in tile configuration."

src/collective/cover/browser/templates/layoutedit.pt
```diff
@@ -51,13 +51,9 @@
                         <div id="slider"></div>
                     </div>
                     <div id="class-chooser" title="Choose Class">
```
hvelarde: "Choose classes"; `title` needs to be internationalized.

src/collective/cover/testing.py
```diff
@@ -104,9 +104,9 @@ def generate_jpeg(width, height):
                 if abs(z) > 2.0:
                     break
                 z = z * z + c
-            r = i % 4 * 64
-            g = i % 8 * 32
-            b = i % 16 * 16
+            r = i % 4 * 64  # noqa: S001
```
hvelarde: don't include this as this bug is already fixed: https://github.com/gforcada/flake8-pep3101/issues/16

what version of `flake8-pep3101` are you using?

src/collective/cover/static/css/layout_edit.css
rodfersou: the simplest solution in this case is to rename the resource name appending the modify date.

src/collective/cover/static/js/layout_edit.js
rodfersou: the simplest solution in this case is to rename the resource name appending the modify date.

src/collective/cover/testing.py
rodfersou: I don't know how to check this out

src/collective/cover/static/js/layout_edit.js
rodfersou: this needs an upgrade step since this file is into js registry

src/collective/cover/tiles/configuration_widgets/cssclasswidget.pt
rodfersou: done

PR #783, opened 2018-05-30
Remove upgrade steps for old, unsupported releases

hvelarde:
Version 1.1b1 is now 3-years old and we are not going to maintain it.


idgserpro:
Should we bump to 2.x like the discussion we had in https://github.com/collective/sc.social.like/pull/162? Or at least a full 1.8 release?

rodfersou:
I don't like to remove these kind of code because they are some "recipes" that we could use in the future.

How about to transform some of them in functions inserted into ``upgrades/__init__.py`` like the ones to cook CSS and JS?

hvelarde:
@rodfersou AFAIK, I'm not removing that code.

rodfersou:
@hvelarde please read again.. I am suggesting to create new functions from what we learn in the code being removed.

hvelarde:
@rodfersou like what?

rodfersou:
@hvelarde a function to update static resource ID for example

rodfersou:
In the near future we will need to do this when start using webpack in cover.

PR #786, opened 2018-06-25
Review multiple class selection when there are many classes

### LINKED ISSUE #785: UI issue: Multiple class select on tiles feature from 1.7b1 doesn't scale beyond 15 css classes
With the introduction of feature #779 The drop down box which used to be a single select is now not a multi-select but a list of checkboxes that is bound by the dimensions. After +/- 15 items depending on your base font size the drop down is cut off by the overlay. 

Test with vanilla collective.cover buildout: Vertical scrolling is activated in the overlay bounds, but the mouse pointer changes as well and I cannot select/activate the scroll bar of the overlay. Work around at the moment is the scroll wheel if your mouse has one. Browsers are not consistent though:

* Chrome Mac I cannot use the overlay scroll bar with the mouse but only use scroll wheel
* Firefox Mac The overlay scroll bar is draggable by the mouse, scroll wheel also scrolls internal overlay
* IE 11 Windows 10, overlay scroll bar is draggable by the mouse, scroll wheel scroll overlay but also the background/page.
* Edge Windows 10, behaves the same as IE11 Windows 10
* Chrome Windows 10 is the worst of both: scroll bar is not draggable and scroll wheel moves the backdrop page instead of the overlay. :-(

To avoid all this: 

I'd expect the scroll-bar to appear on the dropdown menu itself (like the 'native' select) with a max height of 10-12 items too keep it in side the overlay if that's possible. 

chrome_win_10

chrome_mac

rodfersou:
closes #785

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 1.7b3 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- Review multiple class selection when there are many classes (closes `#785 <https://github.com/collective/collective.cover/issues/785>`_).
+  [rodfersou]
+
 - Small code refactor to increase future Python 3 compatibility.
   [hvelarde]
 
--- a/src/collective/cover/browser/templates/layoutedit.pt
+++ b/src/collective/cover/browser/templates/layoutedit.pt
@@ -26,7 +26,7 @@
         <metal:styles fill-slot="style_slot">
             <link rel="stylesheet" type="text/css" media="screen" href=""
                   tal:define="navroot context/@@plone_portal_state/navigation_root_url"
-                  tal:attributes="href string:${navroot}/++resource++collective.cover/css/layout_edit_201804251456.css" />
+                  tal:attributes="href string:${navroot}/++resource++collective.cover/css/layout_edit_201806251610.css" />
         </metal:styles>
         <metal:main fill-slot="main">
             <tal:block tal:define="can_export_layout view/can_export_layout">
--- a/src/collective/cover/static/css/layout_edit_201806251610.css
+++ b/src/collective/cover/static/css/layout_edit_201806251610.css
@@ -741,8 +741,10 @@ body .cssclasswidget-classlist
     z-index: 1001;
 
     display: flex;
+    overflow: auto;
     flex-direction: column;
 
+    max-height: 4rem;
     padding: .4rem;
 
     list-style-type: none !important;
--- a/src/collective/cover/static/css/layout_edit_201806251610.css
+++ b/src/collective/cover/static/css/layout_edit_201806251610.css
@@ -751,7 +751,7 @@ body .cssclasswidget-classlist
     z-index: 1001;
 
     display: flex;
-    overflow: auto;
+    overflow-y: scroll;
     flex-direction: column;
 
     max-height: 4rem;
--- a/src/collective/cover/static/css/layout_edit_201806251610.css
+++ b/src/collective/cover/static/css/layout_edit_201806251610.css
@@ -735,6 +735,16 @@ body .cssclasswidget
 {
     cursor: pointer;
 }
+/* https://stackoverflow.com/a/10100209 */
+body .cssclasswidget-classlist::-webkit-scrollbar {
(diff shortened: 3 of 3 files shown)


CHANGES.rst
```diff
@@ -6,7 +6,8 @@ There's a frood who really knows where his towel is.
 1.7b3 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
-- Nothing changed yet.
+- Review multiple class selection when there is many classes (closes `#785 <https://github.com/collective/collective.cover/issues/785>`_).
```
hvelarde: when there **are** many classes.

hvelarde:
@fredvd would you mind to review this?

hvelarde:
@fredvd any feedback?

fredvd:
@hvelarde.  I’m on vacation this and next week, I’ll take a look and test the fix tomorrow morning.  Thank you for looking into this!

fredvd:
@rodfersou  I've checked out the branch in a local project, the updated css is pulled in, but I don't see any difference in behavior in the css class dropdown:

css-class-dropdown

rodfersou:
@fredvd this is weird.. are you sure you are using this branch version of Cover?

rodfersou:
@hvelarde done

fredvd:
Reviewed on Plain Plone/collective.cover buildout . The problem is fixed from a visual point of view,  as I feared/suspected my own project is overriding some css probably. 

There still is a usability problem now: the drop down selector list doesn't show the scroll bar by default at least in Chrome, so you don't know there are more options and you can click one option and close:  until you are able activate scrolling some way or another in the area (by selecting text and then click scroll, which is non-obvious). 

class-selection_popup-2

rodfersou:
@fredvd the scrollbar just show if there are more items than it can be shown in the predefined space.

rodfersou:
@fredvd we plan to use a plugin instead of this home made widget in the future.

rodfersou:
@fredvd here it works fine
image
I'll take another look in the css..

fredvd:
@rodfersou this was on chrome/mac with 15 configured css classes, only the first 3 show without the scrollbar present on drop down.

rodfersou:
@fredvd it is a bug related to macos; my last commit fixes it.

fredvd:
@rodfersou Testing this at the airport back home now, checked all major browsers on mac (chrome, safari, firefox) just to be sure. You won't believe this: all browsers except for Firefox work now: but with Firefox the scroll bar shows, and then >disappears< after 2-3 seconds. checked firefox devel and firefox stable release 52.9.0. :-(

firefox-mac

Freaking weird. I do get a cross/arrow in the drop down area now after those 3 seconds, so you 'can' see there are more options as soon as you click and the scrollbar is there as soon as you click. So lets keep it like this and merge/close.  Pfew, the old browser incompatibilities from 10 years ago are back in full swing. :-S 

Where is this multi-select widget coming from, from another library, or created by yourself?  All others work, checked Chrom

rodfersou:
@fredvd this one is home made, we plan to use a widget after add webpack in cover buildout.

rodfersou:
@fredvd the disapear behavior happen just on Firefox at Macos; I change a bit the CSS and believe that it is fixed now.. can you test again please?

fredvd:
@rodfersou Back at work. Tested on Firefox mac, works fine now, the scrollbar stays. Tested on all platforms I have access to locally (mac firefox/safari/chrome), (win10: ie11, edge, chrome). Small glitch I saw is that Win-10 IE11 turns the scrollbar into two big arrows to scroll through the list can't remember if it was different before).

But it's totally workable, lets wait for webpack and use an available existing widget where others have done the hard work of fixing browser compatibility. I Never knew a dropdown list could be this complex to function browser independent. 

dropdown_win10-ie11

hvelarde:
@fredvd thanks! do you need now a release for this?

fredvd:
@hvelarde For me it's not super urgent, but others might run into issues with the css drop down as well if they update to 1.7b2 so if no other updates/fixes are planned shortly, better to do a release.

hvelarde:
I made a new release: https://pypi.org/project/collective.cover/1.7b3/

PR #825, opened 2018-09-11
Avoid KeyError when tile schema has changed

hvelarde:
refs. https://github.com/plonegovbr/brasil.gov.portal/issues/524

### DIFF
--- a/src/collective/cover/tiles/data.py
+++ b/src/collective/cover/tiles/data.py
@@ -28,10 +28,14 @@ class PersistentCoverTileDataManager(PersistentTileDataManager):
         conf = self.tile.get_tile_configuration()
         if self.tileType:
             fields = getFields(self.tileType.schema)
-
             for field_name, field_conf in conf.items():
-                if 'order' in field_conf and field_conf['order']:
-                    fields[field_name].order = int(field_conf['order'])
+                if field_name not in fields:
+                    continue  # tile schema has changed
+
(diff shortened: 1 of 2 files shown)


src/collective/cover/tiles/data.py
```diff
@@ -28,10 +28,14 @@ def applyTileConfigurations(self):
...
         if self.tileType:
             fields = getFields(self.tileType.schema)
-
             for field_name, field_conf in conf.items():
-                if 'order' in field_conf and field_conf['order']:
-                    fields[field_name].order = int(field_conf['order'])
+                if field_name not in fields:
+                    continue  # tile schema has changed
+
+                if not isinstance(field_conf, dict):
```
rodfersou: in what case this is not a `dict`?

idgserpro:
Isso não deveria ter sido feito na branch 1.5.x, uma vez que para usar a 2.x, a premissa é atualizar para a última 1.5.x?

hvelarde:
I'm going to do it later: this is an issue on brasil.gov.tiles for IDG 1.2 and we're not responsible for that.

I'm still not pretty sure on how to fix it; we will probably need a monkey patch.

idgserpro:
@hvelarde after run the upgrade, tell us if this did not cause inconsistencies.

hvelarde:
I already did it and I saw no issues: it just skips any field no longer present in the tile schema.

PR #865, opened 2019-09-24
correction for travis tests

Mubra:


idgserpro:
@Mubra I recommend you always create a branch to do a PR and never do in master.

You can do `git reset --hard` on master of your fork, so that it is the same as here again.

Mubra:
@idgserpro  thank you, I thought I had made the branch request, I will keep it in mind

idgserpro:
@Mubra I think it's better to close this PR and create another from a branch other than the master.

Issue #861, opened 2019-09-22
dragged item is copied to all containers (only on previously saved covers)


Mubra:
when a cover is created, saved, and then re-edited, dragged items are copied to all containers

![drag-drop](https://user-images.githubusercontent.com/46692684/65394494-829f9a80-dd54-11e9-8204-e4f414c89e49.gif)

PR #866, opened 2019-09-24
correction travis

Mubra:


idgserpro:
@Mubra  a tip. When creating a branch, name it to reflect what you are doing. Here for example a good name would be `fix_travis`.

@hvelarde  can you invite @Mubra to Collective Organization? I tried but couldn't. Then he can create branches in the `collective/collective.cover` itself.

idgserpro:
@scmos FYI

Mubra:
@idgserpro Good morning, the change in travis, when will it be reflected? Then, because I want to upload the correction of #861  and I know that this change has to be included for the travis tests to work

PR #869, opened 2019-10-01
Solve_issue_868

Mubra:
ambiguity is generated between previously saved items. It was corrected by creating a "row" element and modifying its structure with the even "drop.target" to avoid ambiguity

solve2


Mubra:
hello @idgserpro  ready 👍

idgserpro:
@Mubra please close this PR.

PR #872, opened 2019-10-08
Solve issue 859 and 871

### LINKED ISSUE #859: slide control error
When resizing the tile, generates a javascript error and the control does not slide

image

### LINKED ISSUE #871: Error deleting previously saved items
"x" does not work to delete saved items 

issue_close

Mubra:
The solution was to change the selector "$ le" to "le" because it did not take the charged element, only those created.( closes #871 )
This indirectly   closes #859  using a variable quickly with this selector ("le_resize")

issue_close_final

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,12 @@ There's a frood who really knows where his towel is.
 2.2.1 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- The solution was to change the selector "$ le" to "le" because it did not take the charged element, only those created (fixes `#871 <https://github.com/collective/collective.cover/issues/871>`_).
+    [Mubra]
+      
+  - This indirectly corrects issue `#859 <https://github.com/collective/collective.cover/issues/859>`_ using a variable quickly with this selector ("le_resize") (fixes `#859 <https://github.com/collective/collective.cover/issues/859>`_).
+      [Mubra]
+
 - A "row" object is generated and its location in the DOM is added with the even "drop.target" (fixes `#868 <https://github.com/collective/collective.cover/issues/868>`_).
   [Mubra]
 
--- a/webpack/app/js/layout.js
+++ b/webpack/app/js/layout.js
@@ -9,8 +9,8 @@ export default class LayoutView {
    * @param jqDomObj layout, the layout container
    */
   constructor() {
-    this.$le = $('.layout');
-    this.conf = this.$le.data('layoutmanager-settings');
+    this.le = $('.layout');
+    this.conf = this.le.data('layoutmanager-settings');
     this.n_columns = this.conf.ncolumns;
     this.row_class = 'cover-row';
     this.row_dom = $(
@@ -70,7 +70,7 @@ export default class LayoutView {
     let $button = $('#btn-save');
     let onBtnSave = function(e) {
       e.preventDefault();
-      let json = this.html2json(this.$le);
+      let json = this.html2json(this.le);
       $button.removeClass(function(index, css) {
         return (css.match(/\bbtn-\S+/g) || []).join(' ');
       });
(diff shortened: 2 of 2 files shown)


idgserpro:
@Mubra it's always good to put a `closes #xxx` in the description of a PR to create a link to the issue it fixing. Please add.

Mubra:
ready @idgserpro

CHANGES.rst
```diff
@@ -6,6 +6,12 @@ There's a frood who really knows where his towel is.
 2.2.1 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
+- The solution was to change the selector "$ le" to "le" because it did not take the charged element, only those created (fixes `#871 <https://github.com/collective/collective.cover/issues/871>`_).
```
hvelarde: these changelog entries are wrong; please fix them.

rodfersou:
well, use the `$` in the variable name is a pattern used to know that the variable is a jquery object, and we have it in many projects, please keep this way.

webpack/app/js/layout.js
```diff
@@ -355,12 +356,11 @@ export default class LayoutView {
...
-    let onClick = function() {
-      let element = button.parent('div');
+    button.click(function() {
+      var element = $(this).parent('div');
       element.remove();
-      this.$le.trigger('modified.layout');
-    };
-    button.on('click', onClick.bind(this));
+      le_delete.trigger('modified.layout');
+    });
```
rodfersou: please use the `bind this` in the event instead to fix the this inside the jquery event. ex.:

https://github.com/collective/collective.cover/blob/master/webpack/app/js/cssclasswidget.js#L19

this way you don't need the `le_delete` variable.

also we are using ES6 here, prefer to use `let` and `const` instead of `var` to create variables; prefer to use arrow functions instead of old functions if possible.

please take a looks at http://es6-features.org

webpack/app/js/layout.js
```diff
@@ -478,7 +479,7 @@ export default class LayoutView {
       $('#slider').off("slide");
       $('#slider').on("slide", function(event, ui) {
         column.attr('data-column-size', ui.value);
-        le.trigger('modified.layout');
+        le_resize.trigger('modified.layout');
```
rodfersou: please look at the comments about `le_delete`, the same apply here.

Mubra:
Ok @rodfersou , reviewing, thanks for the comments 👍

webpack/app/js/layout.js
scmos: "le_resize" is needed cause "le" is not reached into function

Mubra:
proof
show

webpack/app/js/layout.js
rodfersou: yes, that's why you should use the `.bind(this)` code when add an event, look at the other comments please.

CHANGES.rst
```diff
@@ -6,6 +6,12 @@ There's a frood who really knows where his towel is.
...
+    [Mubra]
+      
+  - This indirectly corrects issue `#859 <https://github.com/collective/collective.cover/issues/859>`_ using a variable quickly with this selector ("le_resize") (fixes `#859 <https://github.com/collective/collective.cover/issues/859>`_).
```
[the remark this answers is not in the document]
Mubra: I found the problem, if I need to declare "let le_resize = this.$le;". @Rodfersou's comment is correct (that's why I had deleted this variable), but using a variable "le_resize" I think, helps to understand that it is a local object, if you add this, the console error should no longer appear.

CHANGES.rst
Mubra: I don't know if the recent changes in "Travis" affect the tests, I corrected the error, adding these variables to the environment:
    export ZSERVER_PORT = 55001
   export LISTENER_HOST = 49999

With this, the robot tests run me without any problem, in fact I already did some simple tests of mine

CHANGES.rst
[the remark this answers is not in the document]
Mubra: I don't know if the recent changes in "Travis" affect the tests, I corrected the error, adding these variables to the environment:
    export ZSERVER_PORT = 55001
   export LISTENER_HOST = 49999

With this, the robot tests run me without any problem, in fact I already did some simple tests of mine

Mubra:
I will close this PR to generate a new one with the correct changes, also to verify the "travis" tests.
@hvelarde 
@idgserpro 
@rodfersou 
@scmos

PR #875, opened 2019-10-24
Set TZ environment variables in tests to avoid test errors

idgserpro:
`tzlocal` 2.0.0 verifies that the timezone it fouds has the same offset as
the local computer is configured with.If not, it raise the error:

```
ValueError: Timezone offset does not match system offset: -7200 !=
-10800. Please, check your config files.
```

Setting TZ environment variable, this check is no longer done.

Ref: https://github.com/regebro/tzlocal/issues/74

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -79,5 +79,9 @@ recipe = sc.recipe.staticresources
 name = collective.cover
 short_name = cover
 
+[test]
+initialization +=
+    os.environ['TZ'] = 'UTC'
+
 [versions]
 sc.recipe.staticresources = 1.1b5


idgserpro:
@hvelarde this error may be being undercover in Travis due to:

https://github.com/collective/collective.cover/blob/fa35c561bb21ff8cf46cfc367112035a98e9c33f/buildout.cfg#L9

I really don't like `newest = false` in continuous integration.

There are other ways to avoid the restarting buildout, such as using a Python without setuptools/buildout or using requirements.txt.

@mauritsvanrees FYI

PR #881, opened 2020-09-11
Fixed changing the order in a list tile by drag and drop.

mauritsvanrees:
Changing the order in a list tile on the Compose tab would seemingly work at first, but nothing was saved.
The wrong element was taken as base (the whole html document instead of the tile).
And iterating over the children to get their uuids was done wrong, trying to take a uuid attribute from always that same wrong element.

Tested in Plone 4.3 and 5.1.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,8 @@ There's a frood who really knows where his towel is.
 2.2.3 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
-- Nothing changed yet.
+- Fixed changing the order in a list tile by drag and drop.
+  [maurits]
 
 
 2.2.2 (2019-12-27)
--- a/webpack/app/js/compose.js
+++ b/webpack/app/js/compose.js
@@ -148,11 +148,12 @@ export default class ComposeView {
       return;
     }
     let onStop = function(e, ui) {
-      let $el = $(e.currentTarget);
       let uuids = [];
-      $(e.currentTarget).children().each(function(index) {
-        if ($el.attr('data-content-uuid') !== undefined) {
-          uuids.push($el.attr('data-content-uuid'));
+      // We iterate over the children of the original $el determined at the top of onMouseOverSortable.
+      $el.children().each(function(index) {
+        let child = $($el.children()[index]);
+        if (child.attr('data-content-uuid') !== undefined) {
+          uuids.push(child.attr('data-content-uuid'));
         }
       });
       let tile = $el.closest('.tile');


fredvd:
@idgserpro Are you using collective.cover latest versions in production? In this test only the Plone 5.1 build is failing with an obscure version dependency isse where something ',<5.0.0'.    We can look into this, but need cover at the moment for a Plone 4.3 site.

idgserpro:
@fredvd I think it might have to do with Travis' Python setuptools version. Travis configuration uses Python of Travis:

https://github.com/collective/collective.cover/blob/cf134253a7899277c052949bde46d50030b11485/.travis.yml#L33

Perhaps installing the buildout with the pip this problem will be solved.

mauritsvanrees:
I have rebased on the branch from PR #882 and force pushed. That should fix Travis.

mauritsvanrees:
Rebased on master and force pushed for the Travis/QA fixes.

idgserpro:
@mauritsvanrees I saw in the robot tests that checking the ordering of the items is done without reloading the page. See:

https://github.com/collective/collective.cover/blob/980a9bdf04278506a24ca5a0f0bae74825e3d1cd/src/collective/cover/tests/test_list_tile.robot#L88-L110

Perhaps you can reload the page before checking. This could be done in another PR, to confirm that the tests will fail, without fix. Then you can rebase here, to verify that the test will pass.

mauritsvanrees:
I see that on Plone 5 the [robot tests are completely skipped](https://github.com/collective/collective.cover/blob/2.2.2/src/collective/cover/tests/test_robot.py#L34-L36), which is already the case since [2016](f69d6cc6115d6a5aea317dd3660e4bef4b461c3a). When I try them, it quickly fails with `BadRequest: The id "title" is invalid - it is already in use.`.

And with Plone 4.3 I cannot run robot tests, because the remote webdrivers I have available are not compatible with robot framework and friends.

So I cannot really edit robot tests.

fredvd:
@idgserpro  @mauritsvanrees  So to summarise: the robot tests on Plone 4 can no longer work because the selenium webdrivers are no longer functionning and on Plone 5 the robot tests were always skipped already....

I have verified the correct working of the list tile fix from Maurits on Plone 4.  @idgserpro: This was why I asked previously in this thread if you are supporting collective.cover in production installations at the moment. Are these Plone 4 or also Plone 5.0/5.1?  Then we could verify the working of the list/ and related tiles manually and merge this fix.  

I have the impression that collective.cover is just like Plone 4 end of life. We still have to support it on Plone 4 for 1 - 1.5 years but then our maintenance interestes will also be over. We don't have the resources at the moment to fix the robot test infrastructure just to merge this pull request. 

Otherwise we'll just fork and keep our own maintenance branch

hvelarde (approved):
LGTM

hvelarde:
RF tests are running and passing on Travis as you can see here: https://travis-ci.org/github/collective/collective.cover/jobs/727634332

IIRC, there was a problem on supporting both Plone 4 and 5 with RF tests at the same time.

I think the testing matrix can be now simplified and at some point the RF tests must be migrated to support Plone 5.2 only.

collective.cover has been Plone 5.1 compatible since December last year.

mauritsvanrees:
For the record: the problem I have with robot tests on Plone 4.3 is not specific to `collective.cover`. I can't get them to run for the Plone coredev buildout either. I have not tried in a while.

idgserpro:
@mauritsvanrees  I managed to get test robots to work on Plone 4.3, with newer versions of Firefox. See:

https://github.com/plonegovbr/brasil.gov.agenda/commit/cbbc7be296a7d8acc4579d47ee54d8502ba038d4

PR #882, opened 2020-09-15
Travis: use virtualenv and pip to install buildout.

mauritsvanrees:
This might fix Travis on Plone 5.1.
See https://github.com/collective/collective.cover/pull/881#issuecomment-692040474

I triggered a build on current master without changes, and it [failed on 5.1](https://travis-ci.org/github/collective/collective.cover/jobs/727418468)


mauritsvanrees:
@idgserpro You were right with your hunch that using a virtualenv instead of `bootstrap.py` would fix Travis.

buildout.cfg
```diff
@@ -86,3 +86,5 @@ initialization +=
 
 [versions]
 sc.recipe.staticresources = 1.1b5
+setuptools = 42.0.2
+zc.buildout = 2.13.3
```
idgserpro, line 90: Generally I like to pin the versions of the largest Plone supported on cfg of smallest version. So, when new versions are pinned in Plone, the build will fail, indicating that we have to update the versions. For exemple:

https://github.com/collective/collective.nitf/commit/10a3e7e340d4d4fe8863c505c104f139a9b048ea#diff-d3e088c4790fbe53fd3120d360cd08ddR3

.travis.yml
```diff
@@ -30,7 +30,8 @@ before_install:
 install:
 - sed -ie "s#test-4.3#test-$PLONE_VERSION#" buildout.cfg
 - sed -ie "s#versions-4.3#versions-$PLONE_VERSION#" buildout.cfg
-- python bootstrap.py
```
idgserpro, line 33: The `bootstrap.py` file can be removed.

idgserpro (changes requested):

idgserpro:
I think now the `newest = false ` can be removed:

https://github.com/collective/collective.cover/blob/b77afb13e87538fc8f6772d61ea2b0b2c6299c14/buildout.cfg#L9

idgserpro:
Now it isn't necessary to uninstall six of Python of Travis:

https://github.com/collective/collective.cover/blob/cf134253a7899277c052949bde46d50030b11485/.travis.yml#L27-L29

buildout.cfg
mauritsvanrees: Interesting!
I have taken it over in this PR.

.travis.yml
mauritsvanrees: Done

mauritsvanrees:
Thanks for the review. I did the changes.

idgserpro (approved):
LGTM

PR #885, opened 2020-09-26
Update Firefox version on Travis

idgserpro:


hvelarde:
why don't you test with a newer version of Firefox instead?

https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-the-firefox-addon-in-headless-mode

idgserpro:
> why don't you test with a newer version of Firefox instead?

I didn't use the latest version because most addons use the standard version of Travis. I can pin the latest version that exists today. But we will have to someday remember to update this version. Letting Travis update automatically I think it's bad because we can waste time fixs incompatibilities of the new versions. Updates have to be at more widely spaced milestones.

> geckodriver seems to be equivalent to Firefox 60

In fact, Firefox >= 60. I saw in changes that it's okay to use the latest version 0.27.0. I will update the version

hvelarde:
is not clear to me which browser is being used on tests now; remove the geckodriver stuff and test again.

idgserpro:
@hvelarde Firefox 81 is being used see:

https://github.com/collective/collective.cover/pull/885/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R17

This can be seen by the `firefox -v` command:

https://travis-ci.org/github/collective/collective.cover/jobs/732946742#L1849

Firefox is the default browser for tests robots. to see:

https://docs.plone.org/external/plone.app.robotframework/docs/source/happy.html#running-tests-with-a-different-browser

I did not change the configuration of which browser to use.

geckodriver it is necessary to use selenium with newer versions of Firefox. See:

https://stackoverflow.com/questions/43660195/why-firefox-requires-geckodriver

PR #886, opened 2020-10-04
Update versions of setuptools/buildout

idgserpro:
These are the versions used by the new release of Plone 5.1


mauritsvanrees (approved):
LGTM

Issue #891, opened 2021-04-07
Plone5.2 and Python2.7 support


cleberjsantos:

cleberjsantos:
WIP: https://github.com/collective/collective.cover/tree/plone52

wesleybl:
@cleberjsantos thanks for your work! I am interested in getting collective.cover working on Plone 5.2 in the near future. Do you really think it's necessary to maintain Python 2 support? Can I continue your work from copy of your branch?

@hvelarde , are the changes on this branch OK for you?

cleberjsantos:
@wesleybl I not consider it ready for use in Plone 5.2 with Python3, currently we use collective.cover with Plone5.2 and Python 2.7 (Demo. https://idgx.dev/), we chose to run it that way, because some requeriments of collective.cover that are not compatible with Python 3 ( Eg. https://github.com/collective/collective.cover/issues/747) and we lacked time to fix everything.

But my plan is to fix these issues to also migrate everything to Plone5.2 and Python3, and, yes, you can continue contribute, it will be very worthwhile for all :)

hvelarde:
@wesleybl IIRC you have to work over 2 different issues to bring Python 3 compatibility closer; check:

https://github.com/collective/collective.cover/issues?q=is%3Aissue+is%3Aopen+label%3A%22python+3%22

the rest of the code base must already work OOTB.

mauritsvanrees:
I checked out the `plone52` branch, but it fails to startup, complaining about a missing file `src/collective/cover/browser/static/resources.pt`. @cleberjsantos do you have an uncommitted file locally?

mauritsvanrees:
`src/collective/cover/browser/static/` is in the `.gitignore` file, that explains why you did not see that the file needed to be committed.

cleberjsantos:
@mauritsvanrees  This file is generated automatically by  WebPack (https://github.com/collective/collective.cover/blob/plone52/webpack/app/resources.pt).

Eg.

https://github.com/forcontent/idgx.buildout/blob/master/development.cfg#L137
https://github.com/forcontent/idgx.buildout/blob/master/development.cfg#L143

mauritsvanrees:
Ah, right, I can't just use a checkout, I have to run webpack first. Thanks.

mauritsvanrees:
Anyway, a few [dependencies do not work yet on Python 3](#issuecomment-916486741), which is what I want to run it on, so for the moment this is a too big hurdle.

[...]

PR #894, opened 2021-09-24
Code Analysis

wesleybl:
- Black
- flake8
- isort


mauritsvanrees (commented):
Technically this seems okay.
But this makes it extremely hard to ever merge the [Plone 5.2 work](https://github.com/collective/collective.cover/issues/891) that @cleberjsantos is doing. So unless he is fine with your changes, I am against merging it. It can easily be done later.

(Currently I think the chances that we will use cover in 5.2 are very small, and we are more likely to use Volto or Mosaic.)

wesleybl:
@mauritsvanrees I'm thinking of a different approach than @cleberjsantos. I don't think it's interesting to have a release compatible with Plone 5.2 and Python 2. I think the next release should only be compatible with Python 3. I don't think anyone will want to create a new site in Python 2.

I also based on the @hvelarde 's comment, to not continue in Cleber' banch:

https://github.com/collective/collective.cover/issues/891#issuecomment-916486741

I can take some ideas from Cleber and bring it to a new branch.

About Black now, I think the sooner it's done, the less it will "pollute" Blame. If done after Plone 5.2 changes, all those changes will be polluted by Black.

Cleber all right for you?

cleberjsantos:
@wesleybl I think the approach is the same, I don't want to maintain Python 2 compatibility as I also said in the comment, I believe @hvelard didn't look at the commits and not test the branch when he commented, I'm working on it, it's not a stationary branch or outdated ;)

So, as I said before, feel free to help, I think the idea instead of creating alternatives, could be to work on the same alternative, which is, definitely migrate to Python 3, the first step I took was to clean and do work in Plone 5.2, the next step is exactly to move to Python 3.

cleberjsantos (commented):
@wesleybl I think the approach is the same, I don't want to maintain Python 2 compatibility as I also said in the comment, I believe @hvelarde didn't look at the commits and not test the branch when he commented, I'm working on it, it's not a stationary branch or outdated ;)

So, as I said before, feel free to help, I think the idea instead of creating alternatives, could be to work on the same alternative, which is, definitely migrate to Python 3, the first step I took was to clean and do work in Plone 5.2, the next step is exactly to move to Python 3.

mauritsvanrees:
> @mauritsvanrees I'm thinking of a different approach than @cleberjsantos. I don't think it's interesting to have a release compatible with Plone 5.2 and Python 2. I think the next release should only be compatible with Python 3. I don't think anyone will want to create a new site in Python 2.

I agree that creating a new Plone 5.2 site in Python 2 with cover is not a goal.
But if there is no branch that (roughly) supports Plone 5.2 and Python 2, then how are you going to migrate an existing 4.3 site? The standard migration path is: first migrate to 5.2-Py2, and then migrate the database to Python 3.

`collective.exportimport` can be an alternative migration path, but there is no code to migrate tiles yet. (I have some code, but that is only tested on a few Mosaic tiles in 5.2, not for cover.)

mauritsvanrees:
> So, as I said before, feel free to help, I think the idea instead of creating alternatives, could be to work on the same alternative, which is, definitely migrate to Python 3, the first step I took was to clean and do work in Plone 5.2, the next step is exactly to move to Python 3.

This sounds like Cleber would like your help on his 5.2 branch. Running black on master does not help for that. But maybe I am misinterpreting.

I don't mind, I would just like to avoid someone needing to do a difficult and tedious merge later.

cleberjsantos:
@mauritsvanrees  That's right :) your interpretation is correct, I need help in the 5.2 branch to get me even closer to the final goal, which is to have collective.cover working with Python 3.

wesleybl:
I really don't want a Python 2 compatible release. The cover code is too complex and I don't want this backpack.

I don't feel safe with migrating from ZODB Python 2 to ZODB Python 3. I think migrations have to be done "by hand" or with collective.exportimport/transmogrifier.

So my question is: Can we have a Python 3 release only?

@mauritsvanrees @cleberjsantos @hvelarde

cleberjsantos:
@wesleybl If you have an urgent need to use it in a project and don't think it's worth keeping compatibility then I think so, you could create a branch and work on it and then submit a merge request, for eg.

wesleybl:
@cleberjsantos I don't want to do it on a separate big branch. I want to do PRs and merge into master as soon as possible. But for me to do that, I need people to agree that the 3.0.0 release it will be Python 3 only.

Your answer wasn't very straightforward but I understand you don't agree that the 3.0.0 release is Python 3 only. Right?

The issue is not just the code's compatibility with Python 2. There are also the upgrade steps. Are you considering taking the upgrade steps to migrate from Plone 4.3 to Plone 5.2?

hvelarde:
I think this is fine and I think right you can have a release that is working in Plone 5.2 with Python 2 and Python 3 with not too much work ahead if you follow my advice of removing the outdated dependencies.

I spent a lot of time trying to fix the most common issues on Python compatibility and right now you're only stuck with an ugly JS stack that has to be cleaned before you can even think on having a release that works in both Python versions.

after removing those dependencies the code is going to be cleaner and the path will be clearer.

hvelarde (approved):

hvelarde:
@cleberjsantos the problem with your branch is that you are mixing a lot of unrelated changes on the code and are forgetting to take into account some of the consequences that those changes are going to create.

I think is better to divide the work in very small and focused changes that address one issue at a time.

wesleybl:
@hvelarde I think we can enjoy the @cleberjsantos  branch and review/merge as soon as possible. From there we can further separate things. Cleber's changes make sense. I will try to rebase.

hvelarde:
let's review the changes but, as I said previously, changing a lot of unrelated things on the same PR is a recipe for disaster :scream: .

PR #901, opened 2022-02-01
Test with Python 3 on CI

wesleybl:


pbauer (approved):
LGTM. We should also start testing Plone 6 and Python 3.9

wesleybl:
> We should also start testing Plone 6 and Python 3.9

@pbauer unfortunately I don't have time to deal with Plone 6 right now. Thanks for your great work!

PR #911, opened 2022-02-18
Fix WrongContainedType when editing a carousel tile

### LINKED ISSUE #909: WrongContainedType just try to edit a tile carousel
In master, when I try to edit a tile carousel that contains an image and changing some data, I get a WrongContainedType:

```
Wrong contained type

{'e6bbceda1a1040e880c36bffd1c8e621': {'order': 0, 'custom_description': 'mm'}, '42f9180a7d1f41de960054409b629e11': {'order': 1}}
```

wesleybl:
The problem was that the key `order` of the `uuids` field was coming in as an int, when it should actually be a string. To see:

https://github.com/collective/collective.cover/blob/cb13a1c9f7094c237f697c1e1a81bf35ce39feeb/src/collective/cover/tiles/list.py#L37

In the migration to Python 3, the `order` field assignment was changed from:

```python
six.text_type(order)
```

to

```python
safe_unicode(order)
```
To see:

https://github.com/collective/collective.cover/commit/82c873b5bedd5a315149faf4112c244bbafe018a#diff-5736af52418c00117a6e6820d0dad4c838000e46066e775a0811421a2f10ac90L211-R2

However, `safe_unicode(order)` does not transform an int into a string, like `six.text_type(order)`.

Then we use `six.text_type(order)` in the order field assignment.

Fixes #909

### DIFF
--- a/src/collective/cover/tests/test_carousel_tile.py
+++ b/src/collective/cover/tests/test_carousel_tile.py
@@ -106,7 +106,7 @@ class CarouselTileTestCase(TestTileMixin, unittest.TestCase):
 
         self.assertTrue(isinstance(uuids, dict))
         self.assertTrue(len(uuids) == 1)
-        self.assertTrue(uuids[obj1.UID()]["order"] == 0)
+        self.assertTrue(uuids[obj1.UID()]["order"] == u"0")
 
         # now we add a second image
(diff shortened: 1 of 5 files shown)


hvelarde (approved):
LGTM

PR #922, opened 2022-03-11
Shows selected folder path in Content Tree

### LINKED ISSUE #921: Folder path selected in Content Tree is not shown
contentchooser

In the image, I selected the `Folder` folder. See that the `Folder` folder path is not shown. So I can't go back to the root of the portal.

Also see that the "x" that clears the item filter is appearing below the input.

wesleybl:
The fieldset that contains the path receives the `autotoc-section` class, which has the `display: none` property.

So we need to be more specific in the css selector, to put the `display: inline-block` property.

Also fixed the position of the "x" that clears the item filter.

Fixes #921

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ There's a frood who really knows where his towel is.
 3.0.0 (unreleased)
 ^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


wesleybl:
@hvelarde @cleberjsantos @agnogueira Can you take a look here please?

hvelarde (approved):
LGTM

PR #927, opened 2022-04-08
Fix test tile sorting

### LINKED ISSUE #925: List tile sorting not working
On the Compose tab, when I drag an item from a list tile to change the order, the item's new position is not recorded. On the screen the order changes, but if we reload the page, the item returns to its original position.

wesleybl:
In commit

https://github.com/collective/collective.cover/commit/c403cbd8dbdd87d2b1b48aa27ee4e16e92a3a3d2

the function `onStop` of `ComposeView`, has been removed to fix the drag and drop of content from one tile to another. However, it is still needed to save the position of a content, when drag and drop occurs on the same tile, for sorting within the tile.But we do a test, to make sure it will only do something, if the drag and drop is done on the same tile. Otherwise, an error occurs when moving content from one tile to another.

Fixes #925

### DIFF
--- a/src/collective/cover/tests/test_list_tile.robot
+++ b/src/collective/cover/tests/test_list_tile.robot
@@ -95,6 +95,9 @@ Test List Tile
     Drag And Drop By Offset  css=${first_item}  0  580
     Sleep  2s  Wait for reordering to occur
 
+    # Ensures sorting will be saved even after page relead.
+    Reload Page
+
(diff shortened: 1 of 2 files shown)


wesleybl:
Can someone please take a look here?

@hvelarde @rodfersou @cleberjsantos @mauritsvanrees

rodfersou (approved):
LGTM

PR #928, opened 2022-04-08
Remove javascript customization that was making ajax calls synchronous

wesleybl:
Replace `ContentChooser`'s `send` method with a jquery ajax call. The `send` method was causing jquery ajax calls to execute synchronously.


wesleybl:
Can someone please take a look here?

@hvelarde @rodfersou @cleberjsantos @mauritsvanrees

PR #930, opened 2022-07-05
fix "Uncaught TypeError: this.getFolderContents is not a function"

idgserpro:
fix error "Uncaught TypeError: this.getFolderContents is not a function" when trying to open a folder in the navigation tree


wesleybl (approved):
LGTM

PR #933, opened 2023-02-01
Drop python36 support

wesleybl:


pbauer (approved):

CHANGES.rst
```diff
@@ -56,7 +56,7 @@ There's a frood who really knows where his towel is.
 - Remove old upgrade steps.
   [wesleybl]
 
-- Add suport to Python 3.6, 3.7 and 3.8.
+- Add suport to Python 3.7 and 3.8.
```
mauritsvanrees, line 59: Please don't change the history. :-)

mauritsvanrees (changes requested):
One small change requested, see comment. Other than that, it is fine with me to drop 3.6.

Tests fail, but that has nothing to do with this PR.

CHANGES.rst
wesleybl: I changed it because the Python 3.6 add support was in the changes of an ungenerated release. So from a new release point of view, it will be without Python 3.6 support.

Anyway, I undo the change and add a new entry.

mauritsvanrees: Ah, sorry, I missed that.

mauritsvanrees (approved):

